PR #1, opened 2019-10-08
MANTA-4607 Update build image name for manta-buckets-api

kellymclaughlin:
Update the build image name to manta-buckets-api. Also rename some other
references to muskie to facilitate building a new image.


boot/setup.sh
```diff
@@ -37,25 +37,25 @@ function wait_for_resolv_conf {
...
+function manta_setup_buckets_api {
     local num_instances=1
     local size=`json -f ${METADATA} SIZE`
     if [ "$size" = "lab" ]
     then
         num_instances=4
     elif [ "$size" = "production" ]
     then
-	num_instances=16
+    num_instances=16
```
trentm: nit: indent

boot/setup.sh
```diff
@@ -71,39 +71,39 @@ function manta_setup_muskie {
...
-            fatal "could not process $muskie_xml_in to $muskie_xml_out"
-
-        svccfg import $muskie_xml_out || \
-            fatal "unable to import $muskie_instance: $muskie_xml_out"
-        svcadm enable "$muskie_instance" || \
-            fatal "unable to start $muskie_instance"
+        local buckets_api_instance="buckets-api-${ports[i]}"
+        local buckets_api_xml_out=$SVC_ROOT/smf/manifests/buckets-api-${ports[i]}.xml
+        sed -e "s#@@BUCKETS_API_PORT@@#${ports[i]}#g" \
+        -e "s#@@BUCKETS_API_INSECURE_PORT@@#${insecure_ports[i]}#g" \
```
trentm: nit: indent

trentm (dismissed):
I take it this will be a number of steps to getting to a working image, right? E.g. this doesn't handle the sapi_manifests changes, etc.

kellymclaughlin:
> I take it this will be a number of steps to getting to a working image, right? E.g. this doesn't handle the sapi_manifests changes, etc.

Ah good catch. I need to update the location of the config file, but I'm waiting to update the actual contents of the config file template until we change the code as well.

boot/setup.sh
kellymclaughlin: Done

boot/setup.sh
kellymclaughlin: Done

trentm (approved):
LGTM.

PR #4, opened 2019-10-25
MANTA-1936 Rip out syslog logging

kellymclaughlin:
This is part of the work described by RFD 142. It removes the use of syslog logging and instead relies on SMF logging for each instance of the service. The log rotation and upload changes are part of [this](https://github.com/joyent/manta-scripts/pull/4) related PR for the `manta-scripts` repo. 

Currently, I have the `manta-scripts` submodule changed to track the corresponding topic branch in the `manta-scripts` repo, but merging that change to `manta-scripts` and updating the submodule commit accordingly will be done prior to merging this PR. 

### Testing

I have been testing this work by building images and deploying them in my lab system. I have verified that the `buckets-api` instances are logging to their respective SMF log files and that the log files are properly rotated and uploaded each hour.

I have also run the `configure` module test that tests the logging configuration in `buckets-api`. A few test cases had to be updated in response to this change, but after the adjustments they are all passing now.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,7 +10,6 @@
 [submodule "deps/manta-scripts"]
 	path = deps/manta-scripts
 	url = https://github.com/joyent/manta-scripts.git
-	branch = MANTA-1936
 [submodule "deps/eng"]
 	path = deps/eng
 	url = https://github.com/joyent/eng.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,16 +1,16 @@
 [submodule "deps/jsstyle"]
-    path = deps/jsstyle
-    url = https://github.com/joyent/jsstyle.git
+	path = deps/jsstyle
+	url = https://github.com/joyent/jsstyle.git
 [submodule "deps/restdown"]
-    path = deps/restdown
-    url = https://github.com/trentm/restdown.git
+	path = deps/restdown
+	url = https://github.com/trentm/restdown.git
 [submodule "deps/javascriptlint"]
-    path = deps/javascriptlint
-    url = https://github.com/davepacheco/javascriptlint.git
+	path = deps/javascriptlint
+	url = https://github.com/joyent/javascriptlint.git
 [submodule "deps/manta-scripts"]
-    path = deps/manta-scripts
-    url = https://github.com/joyent/manta-scripts.git
+	path = deps/manta-scripts
+	url = https://github.com/joyent/manta-scripts.git
 	branch = MANTA-1936
 [submodule "deps/eng"]
-    path = deps/eng
-    url = https://github.com/joyent/eng.git
+	path = deps/eng
+	url = https://github.com/joyent/eng.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,15 +1,16 @@
 [submodule "deps/jsstyle"]
-	path = deps/jsstyle
-	url = https://github.com/joyent/jsstyle.git
+    path = deps/jsstyle
+    url = https://github.com/joyent/jsstyle.git
 [submodule "deps/restdown"]
-	path = deps/restdown
-	url = https://github.com/trentm/restdown.git
+    path = deps/restdown
+    url = https://github.com/trentm/restdown.git
 [submodule "deps/javascriptlint"]
-	path = deps/javascriptlint
-	url = https://github.com/davepacheco/javascriptlint.git
+    path = deps/javascriptlint
+    url = https://github.com/davepacheco/javascriptlint.git
 [submodule "deps/manta-scripts"]
-	path = deps/manta-scripts
-	url = https://github.com/joyent/manta-scripts.git
+    path = deps/manta-scripts
+    url = https://github.com/joyent/manta-scripts.git
+	branch = MANTA-1936
 [submodule "deps/eng"]
-	path = deps/eng
-	url = https://github.com/joyent/eng.git
+    path = deps/eng
+    url = https://github.com/joyent/eng.git
--- a/lib/configure.js
+++ b/lib/configure.js
@@ -181,7 +181,6 @@ function configureLogging(appName, bunyanCfg, verbose) {
 
     var log = bunyan.createLogger({
         name: appName,
-        level: level,
         streams: streams,
         serializers: restify.bunyan.serializers
     });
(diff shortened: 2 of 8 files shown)


.gitmodules
```diff
@@ -1,15 +1,16 @@
...
-	path = deps/javascriptlint
-	url = https://github.com/davepacheco/javascriptlint.git
+    path = deps/javascriptlint
+    url = https://github.com/davepacheco/javascriptlint.git
 [submodule "deps/manta-scripts"]
-	path = deps/manta-scripts
-	url = https://github.com/joyent/manta-scripts.git
+    path = deps/manta-scripts
+    url = https://github.com/joyent/manta-scripts.git
+	branch = MANTA-1936
```
kellymclaughlin: This is only for testing and will be updated prior to merge after the corresponding `manta-scripts` PR is merged.

lib/configure.js
```diff
@@ -175,124 +174,74 @@ function configure(appName, opts, dtProbes) {
...
+    streams.push({
             level: level,
-            stream: sysl
-        });
+            stream: process.stderr
+    });
 
+    if (level >= bunyan.INFO) {
         /*
          * We want debug info only IFF the request fails AND the configured
```
bahamat: Might as well fix this typo while you're in here?

Line 258/209, `IFF`, should probably be `IF`?

lib/configure.js
```diff
@@ -175,124 +174,74 @@ function configure(appName, opts, dtProbes) {
  * Configure the logger based on the configuration data
  *
  * @param {String} appName: Required. The name of the application
- * @param {Object} bunyanCfg: Required. The bunyan configuration data
- * @param {arrayOfBool} verbose: Required. Array of boolean values that
+ * @param {Object} bunyanCfg: Optional. The bunyan configuration data
```
bahamat, line 131: Why are these now optional? Lines 177,178 in new version.

kellymclaughlin: They were actually always optional, but they had just been mislabeled. You could omit the bunyan config and the logging would just default to `STDERR` and the verbosity flags were usually just for command line invocation when developing or debugging.

.gitmodules
```diff
@@ -1,15 +1,16 @@
...
 [submodule "deps/restdown"]
-	path = deps/restdown
-	url = https://github.com/trentm/restdown.git
+    path = deps/restdown
+    url = https://github.com/trentm/restdown.git
 [submodule "deps/javascriptlint"]
-	path = deps/javascriptlint
-	url = https://github.com/davepacheco/javascriptlint.git
+    path = deps/javascriptlint
+    url = https://github.com/davepacheco/javascriptlint.git
```
trentm: (FYI: "nit" from me can always be ignored. It indicates minor stuff that is unrelated to the important change.)

nit: Perhaps s/davepacheco/joyent while you are at it.

.gitmodules
```diff
@@ -1,15 +1,16 @@
 [submodule "deps/jsstyle"]
-	path = deps/jsstyle
-	url = https://github.com/joyent/jsstyle.git
+    path = deps/jsstyle
+    url = https://github.com/joyent/jsstyle.git
```
trentm: nit: 'sup with the indentation changes. It doesn't break anything, but differs from all our other repos' `.gitmodules` files.  Guessing this is an editor config that changed this?

lib/configure.js
trentm: @bahamat  "IFF" is commonly used in logic/mathematics to mean "if and only if". IOW, I don't believe that is a typo.

lib/configure.js
```diff
@@ -175,124 +174,74 @@ function configure(appName, opts, dtProbes) {
...
-         * If the configured logging level is at or below DEBUG then enable
-         * source logging.
-         */
-        if (level <= bunyan.DEBUG) {
-            log = log.child({src: true});
-        }
-
-        return (log);
+    if (verbose) {
+        level = Math.max(bunyan.TRACE, (level - verbose.length * 10));
```
trentm, line 151: Ha ha, wow. Muskie provides four ways to set the level: default, LOG_LEVEL envvar, `-v` arg(s), config file.

lib/configure.js
```diff
@@ -175,124 +174,74 @@ function configure(appName, opts, dtProbes) {
...
+                }]
+            })
         });
-    } else {
-        log.level(level);
     }
 
+    var log = bunyan.createLogger({
+        name: appName,
+        level: level,
```
trentm: nit: The bunyan docs certainly aren't good at explaining this, but given you've set a level on every stream in `streams`, you don't need to set a `level` at the top-level. In other words, this `level: level,` line could be removed.

trentm (commented):
I think this all looks good. I had questions on the manta-scripts change at https://github.com/joyent/manta-scripts/pull/4#pullrequestreview-308221656 that could perhaps impact the change here.

lib/configure.js
bahamat: TIL. Sorry about that.

This can be reverted then.

test/configure.test.js
```diff
@@ -14,18 +14,10 @@ var configure = require('../lib/configure.js');
...
     var bunyanCfg2 = {
-        level: 'debug',
-        syslog: {
-            facility: 'local0',
-            type: 'udp'
-        }
+        level: 'debug'
     };
 
     var bunyanCfg3 = {
```
bahamat: Seems like `bunyanCfg1` and `bunyanCfg3` are redundant now? Likewise with `bunyanCfg2`/`bunyanCfg4`. It also seems like several of the tests are redundant?

lib/configure.js
kellymclaughlin: The wording is a bit odd to me. I think it's the *only*. I'll reword it a bit.

.gitmodules
kellymclaughlin: Not sure about this. I did `git submodule set-branch` and this is what happened. I'll double-check that though and make sure it's not emacs exerting its will rather than git.

test/configure.test.js
kellymclaughlin: Yes, thanks for catching that. I'll update the tests.

bahamas10 (commented):
This PR has conflicts, but the change itself looks good to me.

bahamas10 (dismissed):
LGTM!

trentm (approved):

bahamas10 (approved):
LGTM!

PR #6, opened 2019-10-30
MANTA-4604 Remove electric-boray from the buckets request path

bahamas10:
This is a combination of changes from @kellymclaughlin and me.

* https://github.com/joyent/manta-buckets-api/pull/6
* https://github.com/joyent/node-boray/pull/11
* https://github.com/joyent/electric-boray/pull/9

### DIFF
--- a/lib/boray_client.js
+++ b/lib/boray_client.js
@@ -51,7 +51,8 @@ function createClient(options, callback) {
         borayargs.srvDomain = pnodeUrl.hostname;
         borayargs.cueballOptions.defaultPort = parseInt(pnodeUrl.port, 10);
         borayargs.log = options.log.child({
-            component: 'boray-client-' + pnodeUrl.hostname
+            component: 'BorayClient',
+            pnode: pnodeUrl.hostname
         });
         borayargs.crc_mode = options.crc_mode;
 
--- a/lib/boray_client.js
+++ b/lib/boray_client.js
@@ -1,0 +1,85 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Copyright 2019 Joyent, Inc.
+ */
+
+var assert = require('assert-plus');
+var clone = require('clone');
+var fs = require('fs');
+var boray = require('boray');
+var url = require('url');
+var verror = require('verror');
+
+/*
+ * Create boray clients in order to interact with boray instances.  Available
+ * boray clients are listed in the ring configuration in LevelDB, which we
+ * access in electric boray via node-fash.
+ */
+function createClient(options, callback) {
+    assert.object(options, 'options');
+    assert.object(options.log, 'options.log');
+    assert.arrayOfString(options.pnodes, 'options.pnodes');
+    assert.object(options.borayOptions, 'options.borayOptions');
+    assert.func(callback, 'callback');
+
+    var log = options.log;
+
+    var clientMap = {};
+    var clientArray = [];
+
+    var pnodes = options.pnodes;
+
+    pnodes.forEach(function (pnode) {
+        var pnodeUrl = url.parse(pnode);
+        assert.string(pnodeUrl.port, 'pnodeUrl.port');
+        assert.string(pnodeUrl.hostname, 'pnodeUrl.hostname');
+
+        log.info({
+            url: pnodeUrl
+        }, 'creating boray client');
+
+        var borayargs = clone(options.borayOptions);
+        if (!borayargs.cueballOptions) {
+            borayargs.cueballOptions = {};
+        }
+        borayargs.unwrapErrors = true;
+        borayargs.srvDomain = pnodeUrl.hostname;
+        borayargs.cueballOptions.defaultPort = parseInt(pnodeUrl.port, 10);
+        borayargs.log = options.log.child({
+            component: 'boray-client-' + pnodeUrl.hostname
+        });
+        borayargs.crc_mode = options.crc_mode;
+
+        var client = boray.createClient(borayargs);
+        clientMap[pnode] = client;
+        clientArray.push(client);
+
+        if (clientArray.length === pnodes.length) {
+            // write ready cookie when clients have connected
+            log.info('all boray clients instantiated writing ready cookie');
+            try {
+                fs.writeFileSync('/var/tmp/electric-boray-ready', null);
+            } catch (e) {
+                throw new verror.VError(e, 'unable to write ready cookie');
+            }
+        }
+    });
+
+    if (clientArray.length <= 0) {
+        throw new verror.VError('No boray clients exist!');
+    }
+
+    return callback(null, {
+        map: clientMap,
+        array: clientArray
+    });
+}
+
+module.exports = {
+    createClient: createClient
+};
--- a/lib/buckets/common.js
+++ b/lib/buckets/common.js
@@ -225,7 +225,7 @@ function _list(type, req, bucket_id) {
                 break;
             }
 
-            ee.emit('entry', obj);
+            ee.emit('entry', obj, record);
         },
         function done(err) {
             if (err) {
--- a/lib/buckets/common.js
+++ b/lib/buckets/common.js
@@ -18,7 +18,7 @@ var format = util.format;
 
 require('../errors');
 
-var BORAY_LIMIT = 1024;
+var LIST_LIMIT = 1024;
 var BORAY_MULTIPLIER = 2;
 
 /*
@@ -73,10 +73,8 @@ function _list(type, req, bucket) {
     assert.object(req.params, 'req.params');
     assert.optionalString(bucket, 'bucket');
 
-    var client = req.boray.client;
     var ee = new EventEmitter();
     var log = req.log;
-    var requestId = req.getId();
 
     var owner = req.owner.account.uuid;
     var prefix = req.params.prefix;
@@ -88,19 +86,20 @@ function _list(type, req, bucket) {
     var limit;
     if (req.params.limit) {
         limit = parseInt(req.params.limit, 10);
-        if (isNaN(limit) || limit <= 0 || limit > 1024) {
+        if (isNaN(limit) || limit <= 0 || limit > LIST_LIMIT) {
             process.nextTick(function () {
                 ee.emit('error', new InvalidLimitError(req.params.limit));
             });
             return (ee);
         }
     } else {
-        limit = 1024;
+        limit = LIST_LIMIT;
     }
 
     assert.number(limit, 'limit');
     assert.ok(limit > 0, 'limit > 0');
-    assert.ok(limit <= 1024, 'limit <= 1024');
+    assert.ok(limit <= LIST_LIMIT,
+        format('limit <= LIST_LIMIT (%d)', LIST_LIMIT));
 
     log.debug('buckets common _list (%s): entered', type);
 
@@ -108,11 +107,13 @@ function _list(type, req, bucket) {
     switch (type) {
     case 'buckets':
         mreq = listBucketsNoVnode(owner, prefix, limit, marker,
-            delimiter, requestId, req.log);
+            delimiter, req);
         break;
     case 'objects':
+            /*
         mreq = client.listObjectsNoVnode(owner, bucket, prefix, limit, marker,
             delimiter, requestId);
+            */
         break;
     default:
         assert.fail('unknown type: ' + type);
@@ -122,6 +123,8 @@ function _list(type, req, bucket) {
     mreq.on('record', function (r) {
         assert.object(r, 'r');
         assert.string(r.type, 'r.type');
+
+        log.error({r: r}, 'mreq.on record');
 
         if (r.type === 'message') {
             mreq.emit('message', r);
@@ -159,9 +162,7 @@ function listBucketsNoVnode(owner, prefix, limit, marker, delimiter, req) {
 
     var ee = new EventEmitter();
     var id = req.getId();
-    var log = log.child({
-        req_id: id
-    });
+    var log = req.log;
 
     if (delimiter && delimiter.length > 1) {
         ee.emit('error', new Error(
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -15,15 +15,18 @@
         "artedi": "1.1.1",
         "assert-plus": "0.1.5",
         "backoff": "2.3.0",
-        "boray": "git+https://github.com/joyent/node-boray.git#master",
+        "bignum": "0.6.0",
+        "boray": "git+https://github.com/joyent/node-boray.git#MANTA-4604",
```
bahamas10: This perhaps should change since `node-boray` has its own PR

kellymclaughlin: Let's just wait and merge the `node-boray`/`node-buckets-mdapi` PR and then update this  right before merging.

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+                }
+                self.log.info('finished boray client connection attempts');
+                self.boray_clients = clients;
+                self.emit('connect');
+            });
+        });
+    });
+
+    this._pollTimer =
+        setTimeout(pollPlacementData.bind(self), self.pollInterval);
```
chudley, line 136: I'm not sure what this means to assign _pollTimer here but also do it inside pollPlacementData(). Which invocation of setTimeout ends up as _pollTimer in this scenario?

I think we could also have pollPlacementData() as a method on MetadataPlacement and just call that method here instead, but I understand not doing that in order to keep the method private. I think that's just the JS world we live in so happy to see it done another way if preferred.

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+var fast = require('fast');
+var jsprim = require('jsprim');
+var once = require('once');
+var util = require('util');
+var vasync = require('vasync');
+var verror = require('verror');
+
+var boray_client = require('./boray_client');
+
+require('./errors');
```
chudley: I feel a bit dumb for this one, but I've not seen a require on its own like this. Is there something in "./errors" that we want to execute but not store here?

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+    });
+
+    this._pollTimer =
+        setTimeout(pollPlacementData.bind(self), self.pollInterval);
+}
+util.inherits(MetadataPlacement, EventEmitter);
+
+
+MetadataPlacement.prototype.close = function close() {
+    clearTimeout(this._storageTimer);
```
chudley: Should this be `_pollTimer`?

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+    if (this.client)
```
chudley: I think this might need to iterate over `electric_boray_client` and `boray_clients` to close each one.

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+
+/**
+ * Function to query electric-boray for updates to the metadata placement
+ * information. This function is called periodically.
+ */
+function pollPlacementData() {
+    assert.object(this.electric_boray_client,
+        'no electric-boray client connected');
+
+    this.log.trace('MetadataPlacement.pollPlacementData: entered');
```
chudley: This message doesn't need to include "MetadataPlacement" because the parent's `log` is a child logger with `component` set.

It looks like this applies to most of the log messages here.

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+        update(tkey).digest('hex');
+    // find the node that corresponds to this hash.
+    var vnodeHashInterval =
+        this.placementData.ring.algorithm.VNODE_HASH_INTERVAL;
+
+    var vnode = parseInt(bignum(value, 16).div(bignum(vnodeHashInterval, 16)),
+        10);
+
+    var pnode = this.placementData.ring.vnodeToPnodeMap[vnode].pnode;
+        var data = this.placementData.ring.pnodeToVnodeMap[pnode][vnode];
```
chudley: Rogue indentation.

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+        data: data
+    });
+};
+
+MetadataPlacement.prototype.getAllNodes = function getAllNodes() {
+    var self = this;
+    var ret = [];
+
+    var map = self.placementData.ring.vnodeToPnodeMap;
+        Object.keys(map).forEach(function (vnode) {
```
chudley: Rogue indentation.

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+    }, 'MetadataPlacement.getObjectLocation: entered');
+
+    var tkey = owner + ':' + bucket + ':' + key;
+
+    log.debug({
+        key: key,
+        tkey: tkey
+    }, 'MetadataPlacement.getObjectLocation: key transformed');
+
+    var value = crypto.createHash(this.placementData.ring.algorithm.NAME).
```
chudley: I think this is done inside `getBucketLocation()` also. It looks like we only do it twice so far, but it's quite a complicated bit of code to read, so would it be possible to lift it into a utility function?

lib/metadata_placement.js
```diff
@@ -0,0 +1,311 @@
...
+    });
+};
+
+
+MetadataPlacement.prototype.toString = function toString() {
+    var str = '[object MetadataPlacement <';
+    str += 'pollInterval=' + this.pollInterval + ', ';
+    if (this.client) {
+        // i.e. NOT initialised in `standalone` mode
+        str += 'moray=' + this.client.toString();
```
chudley: `moray` here is probably going to change with the renaming soon, but thought I'd highlight its use here just in case.

lib/boray_client.js
```diff
@@ -0,0 +1,85 @@
...
+
+        var borayargs = clone(options.borayOptions);
+        if (!borayargs.cueballOptions) {
+            borayargs.cueballOptions = {};
+        }
+        borayargs.unwrapErrors = true;
+        borayargs.srvDomain = pnodeUrl.hostname;
+        borayargs.cueballOptions.defaultPort = parseInt(pnodeUrl.port, 10);
+        borayargs.log = options.log.child({
+            component: 'boray-client-' + pnodeUrl.hostname
```
chudley: I don't think we have a standard for this, but I've mostly seen child loggers set their component to upper camel case.

I think I'd also prefer to see the pnode as its own property on this child logger instead of a suffix of the component name. Not a big deal and I'm not sure whether it'll realistically be used, but it will perhaps bring more flexibility to log searching in the future.

lib/boray_client.js
```diff
@@ -0,0 +1,85 @@
...
+            url: pnodeUrl
+        }, 'creating boray client');
+
+        var borayargs = clone(options.borayOptions);
+        if (!borayargs.cueballOptions) {
+            borayargs.cueballOptions = {};
+        }
+        borayargs.unwrapErrors = true;
+        borayargs.srvDomain = pnodeUrl.hostname;
+        borayargs.cueballOptions.defaultPort = parseInt(pnodeUrl.port, 10);
```
chudley, line 52: I was going to mention an assertion after the `parseInt` usage, but I would expect cueball to also do this so this is fine as it is.

lib/buckets/common.js
```diff
@@ -56,92 +61,186 @@ function listBuckets(req) {
...
     assert.object(req.boray, 'req.boray');
     assert.object(req.params, 'req.params');
-    assert.optionalString(bucket, 'bucket');
+    assert.optionalUuid(bucket_id, 'bucket_id');
 
-    var client = req.boray.client;
     var ee = new EventEmitter();
     var log = req.log;
-    var requestId = req.getId();
+    var id = req.getId();
```
chudley: I think I would prefer `requestId` (or `request_id`) for this variable name. It's a little ambiguous given the other "*id"s in this function, and it's also not used until a fair bit later on.

lib/common.js
```diff
@@ -1777,8 +1777,9 @@ module.exports = {
         function setup(req, res, next) {
             // General request setup
             req.config = options;
-            req.moray = clients.moray;
-            req.boray = clients.boray;
+            // req.moray = clients.moray;
```
chudley: Looks like these could possibly be removed.

lib/buckets/common.js
```diff
@@ -151,3 +250,494 @@ module.exports = {
...
+ *
+ * The `setNewMarker` method can be used to update the marker by throwing out
+ * all records until the new marker is seen.  It is an error to pass a Marker
+ * to this that is less-than (lexicographical sort) the current marker.
+ *
+ * Finally, the `lmstream` object has a `done` property set when the stream has
+ * fully been exhausted.
+ */
+util.inherits(LimitMarkerStream, EventEmitter);
+function LimitMarkerStream(opts) {
```
chudley, line 299: This looks different to the node-lomstream library we have, but do they serve the same purpose?

chudley (commented):
This change is quite big, so hopefully you don't mind me taking a few passes over it!

Some of this work around bucket/object listings (such as `_list` and `LimitMarkerStream`) I think might be lifted from another project into this one, so I haven't reviewed it too heavily as it's likely already received some scrutiny. I would be happy to approve those bits as they are if you're confident that they've already been reviewed as part of integration to the other project.

lib/metadata_placement.js
kellymclaughlin: I have to be honest and say I copied the `picker` module and modified it to suit the needs for metadata polling so I don't know the *why* to all the bits of code here other than "it's what picker does". We're not bound to it so if you think having a method is a better way then that's ok by me.

lib/metadata_placement.js
kellymclaughlin: More copypasta that can probably just be removed in this case.

lib/metadata_placement.js
kellymclaughlin: yes, nice catch!

lib/buckets/common.js
kellymclaughlin: Yep, will change

lib/metadata_placement.js
bahamas10: `./errors.js` does overload `global` with a set of error messages.  It's not at all my favorite pattern, and should probably be refactored to just return the errors as opposed to attaching them to global.

A quick look through this file though shows that none of those exposed errors from `./errors.js` are used here.

lib/buckets/common.js
bahamas10: So, this was originally a LimitOffset (no marker) stream before the listing API was updated a couple of months ago.  I had already written code for that, so I just adapted it to work with a Limit and a Marker.

It probably could be rewritten to use node-lomstream at this point (though I have little experience with that library).  I don't know if they do exactly the same thing however.

bahamas10:
@chudley you are exactly right about `_list` and `LimitMarkerStream` having already been reviewed before they were merged into electric-boray.  The only thing I have added in this review of note is the block comments above `paginationStream` and `LimitMarkerStream` - so I would happily welcome any review on that :)

package.json
kellymclaughlin: Addressed, it's pointing to `master` now which is itself also a temporary condition.

lib/metadata_placement.js
kellymclaughlin: Removed

lib/metadata_placement.js
kellymclaughlin: Fixed

lib/metadata_placement.js
kellymclaughlin: Done

lib/metadata_placement.js
kellymclaughlin: Thanks for pointing that out. I've fixed those up.

lib/boray_client.js
kellymclaughlin: Done

lib/buckets/common.js
kellymclaughlin: Changed

lib/common.js
kellymclaughlin: I removed those two lines.

lib/metadata_placement.js
bahamas10: agreed, will do.

lib/metadata_placement.js
bahamas10: will do

lib/metadata_placement.js
bahamas10: will do

lib/buckets/common.js
chudley: > I don't know if they do exactly the same thing however.

Yea me neither off hand. Seems fine to keep muskie/buckets-api's own implementation if it's already working!

lib/metadata_placement.js
chudley: I think it's probably fine for now given what this change is intended for. As we get more sensitive to data placement changes then perhaps we can revisit to make sure the timing is doing what we expect, but if it's working for picker then I'm happy!

lib/buckets/common.js
```diff
@@ -11,11 +11,16 @@
...
 
 var assert = require('assert-plus');
-var jsprim = require('jsprim');
-var sprintf = require('util').format;
+var util = require('util');
+var vasync = require('vasync');
+
+var format = util.format;
 
 require('../errors');
```
chudley, line 19: Sorry, just noticed this funky import as well.

chudley (dismissed):
Thanks @bahamas10. I checked out the comments you added and they look good.

And thanks both for addressing my feedback! I've had another look over the change as a whole and it looks good to me.

I only have one minor comment that I've added on this latest round, but for the sake of quick iteration I'm happy for it to be addressed in a later change. It won't affect functionality of this service; it's just a little unsightly.

lib/buckets/common.js
bahamas10: This is used to get `InvalidLimitError` defined in this file.  I'll ask about this in ~manta, but I think it would be best to revamp `errors.js` to:

1. not abuse `global`, and
2. not `fs.readFileSync` itself to create exports

as part of a separate PR

kellymclaughlin: +1 to that idea, @bahamas10

chudley (approved):

PR #11, opened 2019-11-06
 MANTA-4615 front door load balancing for buckets

jlevon:
The muppet bits aren't quite ready, but this part is sufficient if you use the MANTA-4615 branch of muppet - maybe we could update the deployment guide to use this repo
instead of the webapi branch, and use the above muppet branch?

I haven't removed the insecure server yet - that can come later.


kellymclaughlin (dismissed):
This looks good and I've tried it out on my lab setup and it's working as expected.

jlevon:
I had a conflict with the resurrection of haproxy (as I removed it) so had to rebase and force push the two csets here. Otherwise it's the same.

kellymclaughlin (approved):

bahamas10:
I haven't tested this specifically, but the code itself LGTM - no obvious issues.

PR #12, opened 2019-11-06
MANTA-4603 Muskie needs deletion of directory operations codebase

kellymclaughlin:
This PR deletes the legacy code from muskie that pertains to directory-based manta operations. This code is not needed as it will live on in the [`manta-muskie`](https://github.com/joyent/manta-muskie) repo.

There are some accompanying changes to the [`sdc-manta`](https://github.com/joyent/sdc-manta) needed also, but a PR is not ready for that yet. I'll update this description with that information once it is available.


bahamas10 (commented):
code itself is :+1: from me - as per our chat i'll wait until the `sdc-manta` changes are made to test this.

kellymclaughlin:
@bahamas10 I've added some testing notes to the Jira ticket and the `sdc-manta` PR is available [here](https://github.com/joyent/sdc-manta/pull/18).

lib/buckets/common.js
```diff
@@ -74,9 +74,11 @@ function _list(type, req, bucket_id) {
         format('invalid _list type: %s', type));
     assert.object(req, 'req');
     assert.object(req.log, 'req.log');
-    assert.object(req.boray, 'req.boray');
     assert.object(req.params, 'req.params');
-    assert.optionalUuid(bucket_id, 'bucket_id');
+    // assert.optionalUuid does not behave properly
```
bahamas10, line 78: This seems like an issue.  Perhaps updating assert-plus to the latest (1.0.0) fixes this?

lib/buckets/list.js
```diff
@@ -84,6 +84,12 @@ function listBuckets(req, res, next) {
 
 function setAuthContext(req, res, next) {
     req.authContext.action = 'getdirectory';
+    var resource = {};
```
bahamas10: nit: any reason to not define the object in one statement? i.e.:

``` js
var resource = {
  key: 'buckets',
  owner: req.owner,
  roles: []
};
...
```

I personally don't mind either way - I'm just curious.

lib/metadata_placement.js
```diff
@@ -250,7 +250,7 @@ MetadataPlacement.prototype.getAllNodes = function getAllNodes() {
         vnode = parseInt(vnode, 10);
         assert.number(vnode, 'vnode');
         var pnode = map[vnode].pnode;
-        assert.number(pnode, 'pnode');
+        assert.string(pnode, 'pnode');
```
bahamas10, line 253: more of a meta question here - should pnode be a number? or a string?

bahamas10 (dismissed):
Changes look good to me. I tested the image in the ticket against my own lab setup (with the latest loadbalancer image as well) and it all seemed to work against some basic buckets requests.

lib/buckets/common.js
kellymclaughlin: It is concerning and updating might fix it. We need to very soon undergo an update of all of our dependencies so I thought that could wait and be part of that effort.

lib/metadata_placement.js
kellymclaughlin: It's a string that looks something like `tcp://2.boray.emy-18.joyent.us:2030`.

lib/buckets/list.js
kellymclaughlin: No, no real reason. I'll make that change.

lib/buckets/common.js
bahamas10: sounds good - I'll look into that with MANTA-4720

lib/metadata_placement.js
bahamas10: Interesting... I don't know how this code didn't cause problems when I tested it then.... weird.  Either way, LGTM

kellymclaughlin: I know the answer to this. It's right [here](https://github.com/joyent/manta-buckets-api/pull/12/files#diff-250badb28fe03ae79416465caa804121L60). A few of us puzzled over this in chat last week and so I decided we'd be better off having the asserts do their job than without.

lib/buckets/list.js
kellymclaughlin: All done

lib/metadata_placement.js
bahamas10: oh wow.  I remember years ago (before I was at joyent) making a PR or an issue for `node-assert-plus` to add "verify" derivatives.  i.e. `assert.verify.bool`, `assert.verify.number`, etc.  where `verify` would always run regardless of an environmental variable being set (this matches the illumos kernel with `ASSERT` and `VERIFY`)... I don't know what happened with that... but NDEBUG for node asserts is really surprising tbh.

bahamas10 (approved):
LGTM!

PR #16, opened 2019-11-15
MANTA-4441 refactored 500 error for invalid delimiter parameter

indianwhocodes:
## Description

Invalid Delimiter Parameter in pagination support leading to Internal Server Error `500`, resolves [MANTA-4441](https://jira.joyent.us/browse/MANTA-4441). 

## Solution

- Performs validation for the condition, when for delimiter if the supplied `string length > 1` char, we return a 400 with the appropriate server message.

### DIFF
--- a/lib/buckets/common.js
+++ b/lib/buckets/common.js
@@ -94,12 +94,13 @@ function _list(type, req, bucket_id) {
     assert.uuid(owner, 'owner');
 
     // Validate optional delimiter
-    if (delimiter && delimiter.length > 1) {
-        ee.emit('error', new Error(
-            '%s: delimiter larger than 1 character: %j',
-            funcname, delimiter));
-        return (ee);
-    }
+        if (delimiter && delimiter.length > 1) {
+            process.nextTick(function () {
+                ee.emit('error', new InvalidParameterError('delimiter',
+                    delimiter));
+            });
+            return (ee);
+        }
 
     // Validate optional limit
     var limit;
--- a/test/configure.test.js
+++ b/test/configure.test.js
@@ -5,7 +5,7 @@
  */
 
 /*
- * Copyright (c) 2019, Joyent, Inc.
+ * Copyright 2019 Joyent, Inc.
  */
 
 var bunyan = require('bunyan');


kellymclaughlin (changes requested):
I tested and this works as expected. My only concern is the composition of the error message. A generated error message might look like this: 
```
{
  "code": "InvalidParameter",
  "message": "\"ab\" is invalid (supplied length > 1) for \"buckets.common._list(buckets)\""
}
```
If a user request has supplied more than 1 query parameter it isn't immediately clear what parameter posed the issue. Perhaps we can have the error state *Delimiter "ab" is invalid*.

Also the details about the internal function (*i.e.* bucket.common._list) that had a problem are not useful to a user. I would say we omit that information from the response since it won't mean anything to them.

indianwhocodes:
That was originally my initial approach to how the error should be reported as well.

lib/errors.js
```diff
@@ -287,11 +287,11 @@ function InvalidUpdateError(k, extra) {
 util.inherits(InvalidUpdateError, BucketsApiError);
 
 
-function InvalidParameterError(k, v) {
+function InvalidParameterError(v) {
     BucketsApiError.call(this, {
         restCode: 'InvalidParameter',
         statusCode: 400,
-        message: '"' + v + '" is invalid for "' + k + '"'
+        message: 'delimiter=' + v + ' is invalid '
```
kellymclaughlin: I wouldn't  hardcode *delimiter* in this function because the function is not specific to just the one delimiter use case. I would instead pass the string composed of `"delimiter " + delimiter into this function. Alternatively you could change this function to accept two parameters with the first being the parameter name and the second being the parameter value and compose the message from those two. I probably prefer the latter option because I think it's more clear.

kellymclaughlin (changes requested):

kellymclaughlin (approved):

lib/buckets/common.js
```diff
@@ -94,12 +94,13 @@ function _list(type, req, bucket_id) {
     assert.uuid(owner, 'owner');
 
     // Validate optional delimiter
-    if (delimiter && delimiter.length > 1) {
-        ee.emit('error', new Error(
-            '%s: delimiter larger than 1 character: %j',
-            funcname, delimiter));
-        return (ee);
-    }
+        if (delimiter && delimiter.length > 1) {
```
bahamas10, line 97: is the indentation modified here?

indianwhocodes: It was.

PR #18, opened 2019-11-20
MANTA-4752 eliminate all #master deps in package.json files

trentm:


lib/metadata_placement.js
```diff
@@ -18,7 +18,7 @@ var EventEmitter = require('events').EventEmitter;
 
 var assert = require('assert-plus');
 var bignum = require('bignum');
-var boray = require('boray');
+var mdapi = require('buckets-api');
```
kellymclaughlin: Shouldn't this also be `require('buckets-mdapi')`?

kellymclaughlin (changes requested):

lib/metadata_placement.js
trentm: ug, sorry about that.

kellymclaughlin (dismissed):

bahamas10 (dismissed):

trentm:
Earlier patches of this would have had the same problem as https://jira.joyent.us/browse/MANTA-4755

kellymclaughlin (approved):

PR #22, opened 2019-11-22
MANTA-4764 remove libmanta dep from buckets-api

bahamas10:
Tested this on my buckets-api zone and ran the node-manta tests and it all works as it should.


lib/auth.js
```diff
@@ -1067,6 +1066,164 @@ function authorize(req, res, next) {
     next();
 }
 
+/*
```
bahamas10: this is *completely* copied & pasted from libmanta

lib/auth.js
```diff
@@ -1067,6 +1066,164 @@ function authorize(req, res, next) {
...
+    assert.arrayOfString(opts.context.conditions.activeRoles,
+        'condition.activeRoles'); // array of UUID
+
+    var mahi = opts.mahi;
+    var action = opts.context.action;
+    var resource = opts.context.resource;
+    resource.roles = resource.roles || [];
+
+    // Authorize if public GET.
+    if (common.PUBLIC_STOR_PATH.test(resource.key) &&
```
bahamas10: we can probably remove this for buckets, right?

kellymclaughlin: yeah

lib/auth.js
trentm: Would it make sense to (a) pull all the libmanta pieces that are used into a separate `./libmantalite.js` to highlight the connection and/or (b) have a comment pointing out the bits that are pulled directly from libmanta?

lib/common.js
```diff
@@ -792,6 +791,37 @@ function getSharkInfo(req, hostname) {
     return (sharks[0]);
 }
 
+/**
+ * Given an resource like `/mark/stor/my%resume.pdf`, this API will convert to
```
trentm: I haven't had a chance to dig in but, I have this sneaking suspicion that this "normalization" is just wrong, at least for buckets API.  Buckets API should *not* be blindly running `decodeURIComponent` on every slash-separated component of the full URL path.  Its individual endpoints should be properly handling encoding/decoding of the separate `login`, `bucketName`, and `objectName` fields, no?

I'm fine if looking into that is separate from this ticket. I gather this is just about moving existing usage over from libmanta.

lib/auth.js
bahamas10: I like the idea of a `libmantalite.js`... i'll make that change.

lib/common.js
bahamas10: I haven't looked into this at all - my goal was just to pull `libmanta` into this file like you mentioned.  I'll create a separate ticket for this.

bahamas10: https://jira.joyent.us/browse/MANTA-4780

bahamas10:
~~fails on node v0.10, works on v6~~

edit: comment this on wrong issue - ignore

trentm (approved):

PR #24, opened 2019-11-26
MANTA-4787 want better buckets-api API documentation

bahamas10:
I personally find it easier to browse just the new documentation at https://github.com/joyent/manta-buckets-api/blob/MANTA-4787/docs/index.md in conjunction with reviewing the diff.

### DIFF
--- a/docs/make-endpoint-toc
+++ b/docs/make-endpoint-toc
@@ -1,0 +1,57 @@
+#!/usr/bin/awk -f
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This is a super basic and super dumb awk script that can be used to make a
+# GitHub compatible Table-of-Contents in Markdown based on API information
+# given in a specific format.  The markdown file must be passed over stdin or
+# as the first argument.  Example:
+#
+# $ ./make-endpoint-toc index.md
+# | name | endpoint |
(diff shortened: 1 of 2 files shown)


docs/make-endpoint-toc
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/awk -f
```
bahamas10, line 1: This file started as an awk-oneliner I used to generate a table of contents based on the API endpoints in the documentation.  I decided, since I thought it was useful, to clean it up and commit it.  I don't expect our API to change *too* much, but I figure this tool maybe useful in the future to automatically parse this endpoint format and create a TOC.

`docs/` may not be the best place for this - I really don't have any emotional investment in this tool either, so i'm open to really anything involving this.

docs/make-endpoint-toc
```diff
@@ -0,0 +1,57 @@
...
+# 1. all endpoint documentation must be contained in an h1 header named
+# "Endpoints"
+# 2. endpoint names begin with an h2 (##) header
+# 3. endpoint paths contain their own code block (4 spaces at the beginning of
+# their line)
+#
+# This script will output a markdown table with links to the h2 (endpoint
+# names) anchors.
+#
+
```
indianwhocodes, line 27: nit: `whitespace`

indianwhocodes (dismissed):

docs/make-endpoint-toc
bahamas10: ah, good catch

indianwhocodes (approved):

PR #32, opened 2019-12-09
MANTA-4822 Uncaught InvalidHeaderError: Invalid Signature Authorization header:

kellymclaughlin:
This is a bug that was inherited from the muskie codebase. Rather than throw in this situation we should just return a 400 error to the caller.

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -900,7 +900,7 @@ module.exports = {
             } catch (e2) {
                 var err = new restifyErrors.InvalidHeaderError(
                     'Invalid Signature Authorization header: ' + e2.message);
-                throw (err);
+                return (next(err));
             }
         }
 


jlevon (approved):
That was simpler than I thought, thanks!
Also, I didn't realise this existed in muskie already, sorry for placing it under buckets.
But it seems low-risk/high-value enough that I might take a look at a mantav1
port of your fix, what do you think?

kellymclaughlin:
> That was simpler than I thought, thanks!
Also, I didn't realise this existed in muskie already, sorry for placing it under buckets.
But it seems low-risk/high-value enough that I might take a look at a mantav1
port of your fix, what do you think?

Agreed, it would be a very minimal change that should be identical to the one here so I think that'd be fine to do.

PR #38, opened 2020-01-01
MANTA-4922 "Uncaught AssertionError: false == true" caused buckets-api to crash when an invalid durability-level is used for PUT

bahamas10:
testing notes in the ticket


lib/uploads/create.js
```diff
@@ -1,419 +0,0 @@
-/*
```
chudley, line 1: Is this some MPU stuff that's being removed?

lib/buckets/objects/create.js
```diff
@@ -129,9 +131,10 @@ function parseArguments(req, res, next)  {
     req._copies = copies;
     req._size = len;
     req.objectId = uuidv4();
-    assert.ok(len >= 0);
-    assert.ok(copies >= 0);
-    assert.ok(req.objectId);
+
+    assert.ok(len >= 0, 'len >= 0');
```
chudley, line 135: Thanks for naming these assertions!

lib/buckets/objects/create.js
```diff
@@ -120,7 +120,9 @@ function parseArguments(req, res, next)  {
     // Next determine the number of copies
     copies = parseInt((req.header('durability-level') ||
                        common.DEF_NUM_COPIES), 10);
-    if (copies < common.DEF_MIN_COPIES || copies > maxObjectCopies) {
+    if (typeof (copies) !== 'number' || isNaN(copies) ||
```
chudley, line 123: If we have `jsprim` available in buckets-api then I think that library might handle some of this. This change itself looks fine though.

Looking at how Muskie handles this for non-bucket objects, it seems to do a similar thing to what buckets objects did: https://github.com/joyent/manta-muskie/blob/962618988753791ed3c2c601b5be6068a9e49341/lib/obj.js#L247-L255

Has this always been a problem in Muskie?

chudley (approved):

lib/uploads/create.js
bahamas10: i don't know the specifics of this - i just know that the file wasn't required by anything but matched the same logic around durability level (that's how i found this file)

lib/buckets/objects/create.js
bahamas10: :+1:

lib/buckets/objects/create.js
bahamas10: 1. it looks like `jsprim.parseInteger` exists - however it can return a `Number` or an `Error`, meaning the `typeof` check will still be necessary to see if it failed.  It may allow us to drop the `isNaN` check however.  If you feel strongly I can make the change, otherwise I think this logic looks sound.

2.  I don't know if it has always been a bug... but it certainly looks like it has.  Unless there's some earlier logic to validate the headers not in that file you linked - i think this has always been broken.

PR #40, opened 2020-01-06
MANTA-4591 Consider storage node file layout changes for manta v2

kellymclaughlin:


lib/shark_client.js
```diff
@@ -129,6 +127,21 @@ function _request(opts, cb) {
     }
 }
 
+/*
+ * Use information about the object to craft the storage path.
+ */
+function storagePath(opts) {
+    var storageLayoutVersion = '/v' + opts.storageLayoutVersion + '/';
+    var owner = opts.creator || opts.owner;
```
trentm: Showing my ignorance: is there really an "creator *or* owner" thing here in the buckets code path?

lib/shark_client.js
```diff
@@ -129,6 +127,21 @@ function _request(opts, cb) {
     }
 }
 
+/*
+ * Use information about the object to craft the storage path.
+ */
+function storagePath(opts) {
+    var storageLayoutVersion = '/v' + opts.storageLayoutVersion + '/';
```
trentm: How about having this throw an exception if storageLayoutVersion is not a known version?

lib/shark_client.js
```diff
@@ -311,8 +324,11 @@ SharkClient.prototype.put = function put(opts, cb) {
     assert.object(opts, 'options');
     assert.optionalNumber(opts.contentLength, 'options.contentLength');
     assert.string(opts.contentType, 'options.contentType');
-    assert.string(opts.objectId, 'options.objectId');
-    assert.string(opts.owner, 'options.owner');
+    assert.uuid(opts.bucketId, 'options.bucketId');
+    assert.uuid(opts.objectId, 'options.objectId');
+    assert.string(opts.objectName, 'options.objectName');
+    assert.string(opts.objectNameHash, 'options.objectNameHash');
+    assert.uuid(opts.owner, 'options.owner');
```
trentm, line 360: Add assert for storageLayoutVersion ?

lib/shark_client.js
```diff
@@ -129,6 +127,21 @@ function _request(opts, cb) {
...
+function storagePath(opts) {
+    var storageLayoutVersion = '/v' + opts.storageLayoutVersion + '/';
+    var owner = opts.creator || opts.owner;
+    var bucketId = opts.bucketId;
+    var objectNameHash = opts.objectNameHash;
+    var objectId = opts.objectId;
+    var objectIdAndHash = objectId + ',' + objectNameHash;
+    var objectIdPrefix = objectId.substr(0, 2);
+
+    return (path.join(storageLayoutVersion, owner, bucketId, objectIdPrefix,
```
trentm: Perhaps explicitly use `'/'` for joining rather than `path.join` which is about OS-indep path separators.

lib/shark_client.js
```diff
@@ -311,8 +324,11 @@ SharkClient.prototype.put = function put(opts, cb) {
     assert.object(opts, 'options');
     assert.optionalNumber(opts.contentLength, 'options.contentLength');
     assert.string(opts.contentType, 'options.contentType');
-    assert.string(opts.objectId, 'options.objectId');
-    assert.string(opts.owner, 'options.owner');
+    assert.uuid(opts.bucketId, 'options.bucketId');
```
trentm, line 356: Perhaps drop the "Options needs: ..." section in the block comment for this function as it is out of date and the asserts self-document it better.

lib/buckets/buckets.js
```diff
@@ -142,6 +146,8 @@ function BucketObject(req) {
     self.bucket_name = req.params.bucket_name;
     if (req.params.object_name) {
         self.name = req.params.object_name;
+        self.name_hash =
+            crypto.createHash('md5').update(self.name).digest('hex');
```
trentm, line 166: You decided on md5 for speed?

lib/utils.js
```diff
@@ -68,6 +68,14 @@ function shuffle(array) {
     return (array);
 }
 
+function splitSubstr(str, len) {
```
trentm: Did you end up not using this function?

kellymclaughlin: No, that's not needed any longer. I'll remove it.

lib/buckets/buckets.js
kellymclaughlin: Speed and it didn't require any other dependencies.

lib/shark_client.js
kellymclaughlin: No, this is a vestige of the muskie codebase. I'll take that out.

lib/shark_client.js
```diff
@@ -129,6 +128,26 @@ function _request(opts, cb) {
     }
 }
 
+/*
+ * Use information about the object to craft the storage path.
+ */
+function storagePath(opts) {
+    if (opts.storageLayoutVersion === 2) {
+        var storageLayoutVersion = '/v' + opts.storageLayoutVersion + '/';
```
trentm: Should drop the trailing '/' now that we are joining with '/' below, otherwise we get a double slash.

lib/shark_client.js
```diff
@@ -5,29 +5,28 @@
...
 
 var assert = require('assert-plus');
 var backoff = require('backoff');
 var KeepAliveAgent = require('keep-alive-agent');
 var once = require('once');
 
 var common = require('./common');
-
-
+var utils = require('./utils');
```
trentm: Unused import, I think.

lib/shark_client.js
```diff
@@ -129,6 +128,26 @@ function _request(opts, cb) {
     }
 }
 
+/*
+ * Use information about the object to craft the storage path.
+ */
```
trentm, line 163: Some comment or documentation on the storage path components should be *somewhere*. Could be on this function.

- why the hash
- doc that the hash is an md5
- why "2" for the objectIdPrefix

Having that info be a comment on this function would be reasonable.

lib/shark_client.js
```diff
@@ -129,6 +128,26 @@ function _request(opts, cb) {
...
+function storagePath(opts) {
+    if (opts.storageLayoutVersion === 2) {
+        var storageLayoutVersion = '/v' + opts.storageLayoutVersion + '/';
+        var owner = opts.owner;
+        var bucketId = opts.bucketId;
+        var objectNameHash = opts.objectNameHash;
+        var objectId = opts.objectId;
+        var objectIdAndHash = objectId + ',' + objectNameHash;
+        var objectIdPrefix = objectId.substr(0, 2);
```
trentm: Perhaps pull this `2` out to a constant? Or at the least a comment here or on this function for why "2" here.

trentm (changes requested):

lib/buckets/buckets.js
```diff
@@ -137,6 +137,22 @@ function Bucket(req) {
 
 }
 
+/**
+ * Create a new BucketObject
+ *
+ * Please note the presence of the name_hash field. This field represents the
+ * MD5 hash of the object name. It is used as input to determine where to place
+ * an object's metadata record. This decision to use this value rather than the
+ * raw object name as input the to metadata location hash function was made in
```
trentm:
```suggestion
 * raw object name as input to the metadata location hash function was made in
```

lib/buckets/buckets.js
```diff
@@ -137,6 +137,22 @@ function Bucket(req) {
...
+ * order to allow us to construct the storage file path such that all the inputs
+ * the the metadata placement hash function are present on the storage node and
```
trentm:
```suggestion
 * to the metadata placement hash function are present on the storage node and
```

lib/shark_client.js
```diff
@@ -128,18 +130,46 @@ function _request(opts, cb) {
     }
 }
 
-/*
+/**
  * Use information about the object to craft the storage path.
+ *
+ * An important detail to note is the use of the MD5 hash of the object name
+ * when constructing the storage node file path rather than the actual name of
+ * the object. Metadata placement uses the MD5 has of the object name as input
```
trentm:
```suggestion
 * the object. Metadata placement uses the MD5 hash of the object name as input
```

lib/shark_client.js
```diff
@@ -128,18 +130,46 @@ function _request(opts, cb) {
...
+ * to determine where to place an object's metadata record. This decision was
+ * made in order to allow us to construct the storage file path such that all
+ * the inputs the the metadata placement hash function are present on the
```
trentm:
```suggestion
 * the inputs to the metadata placement hash function are present on the
```

test/storage-path.test.js
```diff
@@ -0,0 +1,63 @@
...
+            objectNameHash: objectNameHash,
+            objectId: uuidv4(),
+            storageLayoutVersion: 2
+        };
+
+        var storagePath = shark_client.storagePath(storagePathOpts);
+
+        var actualPathParts = storagePath.substr(1).split('/');
+        var actualPathPartCount = actualPathParts.length;
+        var expectedObjectIdPrefix = storagePathOpts.objectId.substr(0, 2);
```
trentm: nit: Could use the const for this `2`, or just add a comment with the const name if exporting it is a pain just for testing.

trentm (dismissed):
Just nits.

test/storage-path.test.js
kellymclaughlin: Done

trentm (approved):

PR #41, opened 2020-01-07
MANTA-4946 Specify service name for buckets-mdplacement cueball configuration

kellymclaughlin:
This change overrides the [default service name](https://github.com/joyent/node-buckets-mdapi/blob/master/lib/client_params.js#L39) in `node-buckets-mdapi` so that the SRV record queries for `buckets-mdplacement` are issued properly. Without this change `buckets-api` ends up falling back to A record queries which work, but is not the intended mode of operation.


jlevon (approved):
Some digging revealed why this worked for me previously, but is now necessary. Looks good, thanks!

PR #42, opened 2020-01-14
MANTA-4780 buckets-api libmanta normalizeMantaPath may need to be updated for buckets

bahamas10:
This was originally mentioned by @trentm on https://github.com/joyent/manta-buckets-api/pull/22#discussion_r350329741

This change ends up making `normalizeMantaPath` also act as a form of validation in a way (which should be unnecessary as restify will handle this for us).

NOTE: my implementation of this is very verbose, and perhaps does too much validation / could break in ways I haven't imagined.  For what this does, it works as expected.


trentm:
Is this for https://jira.joyent.us/browse/MANTA-4780 ?

bahamas10:
> Is this for https://jira.joyent.us/browse/MANTA-4780 ?

@trentm yes! sorry about that, i must have pasted over the ticket number when making the title.  I've since updated it.

lib/libmantalite.js
```diff
@@ -19,34 +19,86 @@ module.exports.normalizeMantaPath = normalizeMantaPath;
...
+
+        assert.equal(parts[0], '', 'blank first element');
+
+        // account
+        if (parts[1]) {
+            parts[1] = decodeURIComponent(parts[1]);
+        }
+
+        if (parts[2]) {
+            assert.equal(parts[2], 'buckets');
```
kellymclaughlin: I worry about hardcoding this in a handler that's used for every request. I think could be deferred to the route matching done by restify.

lib/libmantalite.js
```diff
@@ -19,34 +19,86 @@ module.exports.normalizeMantaPath = normalizeMantaPath;
...
+        }
+
+        // bucket name
+        if (parts[3]) {
+            parts[3] = decodeURIComponent(parts[3]);
+        }
 
-        if (opts.account)
+        if (parts[4]) {
+            assert.equal(parts[4], 'objects');
```
kellymclaughlin: Same goes for this as the comment about the 'buckets' check above.

kellymclaughlin (commented):
I'm not completely clear on the problem with the current version of `normalizeMantaPath` that is being solved here. One concern I have is that that it makes an assumption about the shape of the request path where the old code is more general by using `map` . If we added some other set of routes to this application that had more than 5 path components then this might failed to properly decode some of the parts.

lib/common.js
```diff
@@ -939,6 +939,7 @@ module.exports = {
             }
 
             var _opts = {
+                log: req.log,
```
trentm: Correct me if I'm wrong: all we are doing with `libmantalite.normalizeMantaPath` usage is setting `req.key`?
And req.key is used exactly once here: https://github.com/joyent/manta-buckets-api/blob/3b9fa981dff336159989f16bed0789149ed8b8b8/lib/buckets/list.js#L21  in a log statement. IIUC it is a relic of *moray* usage where a moray object has a "key" field as its unique key.  Now we have a SQL "PRIMARY KEY" (https://github.com/joyent/manta-buckets-mdapi/blob/master/schema_templates/schema.in#L52).

If true, I think the normalization func and usage can be dropped.

trentm (changes requested):

trentm:
Also a couple nits:

- The original mention at https://github.com/joyent/manta-buckets-api/pull/22#discussion_r350329741 should be linked in the ticket (MANTA-4780), as it provides some context.
- The ticket should be assigned and in progress.

bahamas10:
@trentm original comment added to the ticket, ticket assigned to me.

lib/common.js
bahamas10: so, hm. you are absolutely right.  I thought I remembered this key specifically in the listing logic, but i'm remembering `r.key` where `r` is from `mreq.on('record', ...)`.... and even then that was swapped for `r.name`.

I just checked on my lab machine inside `/opt/smartdc/buckets-api` to ensure that node_modules was checked as well, and I don't think `req.key` is used in any capacity.  I'll drop it and do some testing, good find!

bahamas10:
@kellymclaughlin @trentm code completely ripped out.  Thank you @trentm for noticing this - i feel a bit silly not catching this myself, but either way It's good to have negative diffs :).

I've tested this code (build `3cd675d3-95f7-446f-8ddb-0d194748b56a` from [jenkins](https://jenkins.joyent.us/job/joyent-org/job/manta-buckets-api/job/PR-42/2/console)) on my lab machine and ran the node-manta tests against it without issue.

trentm (approved):

PR #43, opened 2020-01-15
MANTA-4944 buckets-api: Should do input validation for unhandled characters

bahamas10:
testing notes in the ticket

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -9,7 +9,7 @@ markdown2extras: wiki-tables, code-friendly
 -->
 
 <!--
-    Copyright 2019 Joyent, Inc.
+    Copyright 2020 Joyent, Inc.
 -->
 
 # Manta Buckets Storage Service
@@ -283,8 +283,8 @@ On success this will return a `204` status code with no data.
 
 ### Restrictions
 
-Bucket names must be between 3 and 63 characters long, and must not
-"resemble an IP address" as defined below.
+Bucket names must be between 3 and 63 characters long (calculated in bytes),
+and must not "resemble an IP address" as defined below.
 
 A valid bucket name is composed of one or more "labels" separated by periods.
 
@@ -459,11 +459,11 @@ Also, it will include the following headers:
 
 There are very few limitations imposed on object names. Object names must
 contain only valid UTF-8 characters and may be a maximum of 1024 characters in
-length. Object names may include forward slash characters (or any other valid
-UTF-8 character) to create the suggestion of a directory hierarchy for a set of
-object even though the buckets system uses a flat namespace. Care must be
-taken, however, to properly URL encode all object names to avoid problems when
-interacting with the server.
+length (in bytes). Object names may include forward slash characters (or any
+other valid UTF-8 character) to create the suggestion of a directory hierarchy
+for a set of object even though the buckets system uses a flat namespace. Care
+must be taken, however, to properly URL encode all object names to avoid
+problems when interacting with the server.
 
 ---
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -283,8 +283,8 @@ On success this will return a `204` status code with no data.
 
 ### Restrictions
 
-Bucket names must be between 3 and 63 characters long (calculated in bytes),
-and must not "resemble an IP address" as defined below.
+Bucket names must be between 3 and 63 characters long, and must not
+"resemble an IP address" as defined below.
 
 A valid bucket name is composed of one or more "labels" separated by periods.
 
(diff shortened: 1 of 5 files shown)


lib/buckets/common.js
```diff
@@ -53,8 +58,21 @@ var resemblesIpRegexStr = format('^%s\.%s\.%s\.%s$', threeDigitRegexStr,
...
+}
+
+/*
+ * A valid bucket object name contains characters with valid UTF-8 characters
+ * (non-null) and is a maximum of 1024 characters.
+ */
+function isValidBucketObjectName(name) {
+    assert.string(name, 'name');
+
+    return (!nullRegex.test(name) && name.length > 0 && name.length <= 1024);
```
trentm: - Can we pull this 1024 max len out to a constant?

- Do we want to consider copying S3 which specifies the max length is 1024 *bytes*? Limiting on bytes means you have a physical limit (if theoretically we were calculating DB space requirements). A unicode char, practically speaking, can be 4 bytes in UTF-8, so a 1024 *char* limit would mean a 4096 byte potential upper limit.


```javascript
$ node
> s1 = 'hi'
'hi'
> s2 = '\u{1F436}\u{1F436}'
'🐶🐶'
> s1.length
2
> s2.length
4
> Buffer.byteLength(s1)
2
> Buffer.byteLength(s2)
8
```

If we copy S3, we should include a ref to their docs: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys

lib/buckets/common.js
```diff
@@ -53,8 +58,21 @@ var resemblesIpRegexStr = format('^%s\.%s\.%s\.%s$', threeDigitRegexStr,
 var resemblesIpRegex = new RegExp(resemblesIpRegexStr);
 
 function isValidBucketName(name) {
-    return bucketRegex.test(name) && !resemblesIpRegex.test(name) &&
-        name.length >= 3 && name.length <= 63;
+    assert.string(name, 'name');
```
trentm, line 65: I know you didn't add the sizes here, but it would be good to have some docs on these values. I assume they are coming from S3: https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules and DNS requirements.

trentm (changes requested):

lib/buckets/common.js
bahamas10: There are based on S3, but are documented more thoroughly here: https://github.com/joyent/manta-buckets-api/blob/master/docs/index.md#restrictions

lib/buckets/common.js
bahamas10: - Agreed, I can pull out 1024 and 63 above and make them constants so they aren't as hidden
- Our documentation (linked above) currently states "characters" and not "bytes", so at the very least our documentation is inline with our current implementation.  However, I personally lean more towards preferring `Buffer.byteLength(s)` as opposed to `s.length` as using byte size is less surprising when it comes to actually storing the data.

@kellymclaughlin do you have thoughts on this? I'm leaning towards changing our restrictions to be bytes and not character length.

bahamas10 (commented):
ty for the review

lib/buckets/common.js
kellymclaughlin: Yeah, I agree. I think that's a good idea.

docs/index.md
```diff
@@ -459,11 +459,11 @@ Also, it will include the following headers:
 
 There are very few limitations imposed on object names. Object names must
 contain only valid UTF-8 characters and may be a maximum of 1024 characters in
-length. Object names may include forward slash characters (or any other valid
-UTF-8 character) to create the suggestion of a directory hierarchy for a set of
-object even though the buckets system uses a flat namespace. Care must be
-taken, however, to properly URL encode all object names to avoid problems when
-interacting with the server.
+length (in bytes). Object names may include forward slash characters (or any
```
trentm: s/1024 characters in length (in bytes)/1024 bytes/   or similar.

lib/buckets/common.js
```diff
@@ -53,8 +62,24 @@ var resemblesIpRegexStr = format('^%s\.%s\.%s\.%s$', threeDigitRegexStr,
...
+
+    return (bucketRegex.test(name) && !resemblesIpRegex.test(name) &&
+        !nullRegex.test(name) &&
+        Buffer.byteLength(name) >= MIN_BUCKET_NAME_BYTE_SIZE &&
+        Buffer.byteLength(name) <= MAX_BUCKET_NAME_BYTE_SIZE);
```
trentm: Actually AFAIK the *bucket name* length restriction *is* about characters and not bytes. That's per https://tools.ietf.org/html/rfc1035#section-2.3.1

```
The labels must follow the rules for ARPANET host names.  They must
start with a letter, end with a letter or digit, and have as interior
characters only letters, digits, and hyphen.  There are also some
restrictions on the length.  Labels must be 63 characters or less.
```

trentm (changes requested):

docs/index.md
bahamas10: will do

lib/buckets/common.js
bahamas10: Interesting... will make the change.

bahamas10:
latest image tested with the node-manta tests... all working for me.

trentm (approved):

PR #45, opened 2020-01-29
MANTA-5010 Update buckets-api to use node-fast 3.0.0

kellymclaughlin:


package.json
```diff
@@ -16,15 +16,15 @@
         "assert-plus": "^1.0.0",
         "backoff": "2.3.0",
         "bignum": "0.13.0",
-        "buckets-mdapi": "git+https://github.com/joyent/node-buckets-mdapi.git#v0.4.0",
+        "buckets-mdapi": "git+https://github.com/joyent/node-buckets-mdapi.git#3cb01b4c4a3051ee723401763b725c8c43a39b88",
```
trentm: Were you planning to update this with a v0.5.0 tag?

trentm (dismissed):

package.json
kellymclaughlin: Yep, I'll do that after merging.

trentm (approved):

PR #46, opened 2020-01-31
MANTA-4988 Want mlocate to work with buckets

isaacdavis:
See ticket for testing notes.

This is different enough from the old mlocate that it's probably best just to read the file rather than look at the diff.

You should be able to hotpatch the changes into your own buckets-api zone and have it work as well.


kellymclaughlin (approved):
Test and works as expected. LGTM! Nice job!

PR #50, opened 2020-02-27
MANTA-5063 buckets-api LimitMarkerStream unhandled error causes crash

bahamas10:
testing notes in the ticket https://smartos.org/bugview/MANTA-5063


kellymclaughlin (approved):
Nice work!

PR #53, opened 2020-03-18
MANTA-4341 Optimize conditional object operations

chudley:
This change now leverages the buckets-mdapi service to do _most_ of the conditional request evaluation. As such, lib/conditional_request.js has been removed and replaced with some custom functionality to handle header parsing into a `req.conditions` object, as well as some post-request status code translation in the event of a `304 Not Modified` (which is the rest of the conditional evaluation I hinted at earlier).

My biggest concern with this is that a `CreateBucketObject` request will still have to perform two separate calls to the buckets-mdapi service, each of which subject to the client's conditions.

The request path basically goes like: 

1. `rpc: getobject`
1. Stream object data to storage,
1. `rpc: createobject`

If we didn't do that initial `getobject`, we could end up streaming a large amount of data from the client and only evaluating whether the conditions are satisfied at the end (`createobject`), which wastes resources and creates cruft. Instead, we subject the request to two separate evaluations of the conditions, the second of which could have changed since the initial `getobject`.

I don't think this affects correctness, but it does still represent a case where a request could stream a large amount of data to storage, only to be told that the request has failed on conditions.


lib/buckets/objects/get.js
```diff
@@ -5,25 +5,105 @@
...
+ *     - If-Match
+ *     - If-Unmodified-Since
+ *
+ * The remainder of the supported conditional headers are as follows and are
+ * handled here in buckets-api:
+ *
+ *     - If-None-Match
+ *     - If-Modified-Since
+ *
+ * The reason for this is because in the case of precondition failre in the
```
kellymclaughlin: nit: typo on *failure*

lib/buckets/objects/head.js
```diff
@@ -10,29 +10,48 @@
...
+ *     - If-Match
+ *     - If-Unmodified-Since
+ *
+ * The remainder of the supported conditional headers are as follows and are
+ * handled here in buckets-api:
+ *
+ *     - If-None-Match
+ *     - If-Modified-Since
+ *
+ * See lib/bucekts/objects/get.js for the reasoning behind this.
```
kellymclaughlin: nit: typo on *buckets*

package.json
```diff
@@ -16,7 +16,7 @@
         "assert-plus": "^1.0.0",
         "backoff": "2.3.0",
         "bignum": "0.13.0",
-        "buckets-mdapi": "git+https://github.com/joyent/node-buckets-mdapi.git#v0.5.0",
+        "buckets-mdapi": "git+https://github.com/chudley/node-buckets-mdapi.git#dev-MANTA-4341",
```
kellymclaughlin: I approved the `node-buckets-mdapi` PR so just a note to update this once that's merged.

lib/buckets/buckets.js
```diff
@@ -265,9 +272,54 @@ function createObjectMetadata(req, type, cb) {
...
+            if (client_etag === '*' || client_etag === object_etag) {
+                code = 304;
+                return;
+            }
+        });
+    }
+
+    if ((if_modified_since = req.conditions['if-modified-since'])) {
+        /*
+         * XXX What do we do here if for some reason buckets-mdapi has passed us
```
kellymclaughlin: That's a good question. In normal circumstances this should not ever happen based on the postgres column type used and the serialization in `buckets-mdapi` so if this fails maybe the right thing is to return a 500 to indicate something has really gone wrong?

kellymclaughlin (changes requested):
This is really nice work! Just a few minor comments on it.

lib/buckets/buckets.js
chudley: Actually it looks like this assumption is already made when we initially set the headers (see [here](https://github.com/joyent/manta-buckets-api/blob/0310aa0c4b4f1bc480f7b4c8b5a1e6d78cec2f8b/lib/buckets/objects/metadata/update.js#L72) or [here](https://github.com/joyent/manta-buckets-api/blob/0310aa0c4b4f1bc480f7b4c8b5a1e6d78cec2f8b/lib/buckets/objects/head.js#L69) for examples). The reason this method is doing another parse of the date is because when setting the "Last-Modified" header in the above cases it's cast as a string, so we must re-parse it.

I've removed the `XXX` comment for now, but I'd still like to handle this case while setting the headers. There were a few minor inconsistencies that I started ironing out as part of MANTA-4556 (or as part of the followup I mentioned in the ticket), so I'd be happy to add this scenario to that ticket if that works for you.

kellymclaughlin (approved):

PR #54, opened 2020-03-30
MANTA-5129 mlocate does not seem to work for bucket objects

isaacdavis:
Quick fix -- had to update mlocate to use the object hash to match the changes from MANTA-4591.

See ticket for testing notes.


kellymclaughlin (approved):
Thanks, @isaacdavis!

PR #58, opened 2020-05-01
MANTA-5219 Configure metricPorts in buckets API components by default

askfongjojo:
Tested the change by reprovisioning a bucket-api instance in a lab environment to the image with this PR and confirming that mdata-get metricPorts return the correct values.

### DIFF
--- a/boot/setup.sh
+++ b/boot/setup.sh
@@ -7,7 +7,7 @@
 #
 
 #
-# Copyright 2019 Joyent, Inc.
+# Copyright 2020 Joyent, Inc.
 #
 
 set -o xtrace
--- a/boot/setup.sh
+++ b/boot/setup.sh
@@ -60,9 +60,11 @@ function manta_setup_buckets_api {
 
     #Build the list of ports.  That'll be used for everything else.
     local ports
+    local metric_ports
     local portlist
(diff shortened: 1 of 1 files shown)


boot/setup.sh
```diff
@@ -87,6 +89,19 @@ function manta_setup_buckets_api {
...
     done
 
+    #
+    # We join the metric ports in a comma-separated list, then add this list as
+    # metricPorts mdata to allow scraping by cmon-agent.
+    #
+    # The metricPorts values are derived from the buckets-api service's "SIZE"
+    # SAPI metadata. We don't need to worry about keeping the metricPorts updated
+    # if this variable changes, because such a change does not affect
+    # already-provisioned zones. This is because electric-moray zones pull the
```
isaacdavis: Small nit: change "electric-moray" to "buckets-api". This isn't your fault -- I just looked and the muskie code also has this typo . . .

isaacdavis (changes requested):

boot/setup.sh
askfongjojo: ugh I missed that, fixed now.

isaacdavis (approved):
Looks good; thanks!

PR #64, opened 2022-04-07
TOOLS-2525 Everything needs to stop cloning with git:// URLs (deps)

bahamat:


danmcd (approved):
Thanks for the package.json update! :)

PR #69, opened 2025-12-05
MANTA-5485 Add STS/IAM support to manta-buckets-api 

cneira:
This PR adds support for Secure Token Service (STS) and Identity and Access Management (IAM) to MANTA-5471.
The work done here supersedes MANTA-5482 (https://github.com/TritonDataCenter/manta-buckets-api/pull/68)

 For Security Token Service (STS) operations the following endpoints have been added:
  - AssumeRole: Generate temporary credentials with assumed role
  - GetSessionToken: Create session-scoped temporary credentials
  - GetCallerIdentity: Return details about IAM entity

  For Identity and Access Management (IAM)  the following endpoints have been added:
  - CreateRole/GetRole/DeleteRole: Role lifecycle management
  - ListRoles: Enumerate roles in account
  - PutRolePolicy/GetRolePolicy/DeleteRolePolicy: Inline policies
  - ListRolePolicies: List policies attached to roles
  
    Co-Authored-By: Claude <noreply@anthropic.com>


README.md
```diff
@@ -105,6 +230,286 @@ historical reasons (both muppet and the old webapi registered themselves with a
 service name of "manta"), and to reduce load on
 [binder](https://github.com/TritonDataCenter/binder/).
 
+## S3 Compatibility
+
+Manta Buckets API provides S3-compatible endpoints that translate S3 API requests into Manta bucket operations. This compatibility layer enables S3 clients and tools to work with Manta's bucket storage.
+[!NOTE]
+Only Path-Style URL is supported.
+
```
danmcd, line 241: Drop a mention up here about the compatibility matrix (linked in the `## Documentation` section) for those who visit here expecting to see minutia.

lib/anonymous-auth.js
```diff
@@ -0,0 +1,630 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Copyright 2025 Edgecast Cloud LLC.
+ * File:     anonymous-auth.js
```
danmcd, line 13: Editorial nit:  Split the copyright line into its own block.

danmcd (commented):
Checkpoint, continue with lib/anonymous-auth.js modulo incrementals.

lib/anonymous-auth.js
```diff
@@ -0,0 +1,630 @@
...
+        login: 'anonymous',
+        isAdmin: false
+    }),
+    user: Object.freeze({
+        uuid: 'anonymous',
+        login: 'anonymous'
+    }),
+    roles: Object.freeze(['public-read']), // Anonymous users have public-read
+    // NOTE: Do NOT set anonymous: true here, as that triggers old
+    // Manta anonymous logic
```
danmcd, line 86: Pardon my naivete.  Can we have a Manta deployment that has classic Manta anonymous views (like our `/Joyent_Dev/public`) AND S3-style anonymous user access? Not a blocker, but I need to make sure I didn't miss something and "GO READ xxx" is a perfectly sane answer.

lib/auth/authorization-handler.js
```diff
@@ -0,0 +1,503 @@
...
+                next(new AuthorizationError(login, req.path(), e));
+            } else {
+                next(new NoMatchingRoleTagError());
+            }
+            return;
+        case 'InvalidRole':
+            next(new InvalidRoleError(e.message));
+            return;
+        case 'CrossAccount':
+            /* This should never happen. */
```
danmcd: Why should this never happen? Because it was added per above ('NoMatchingRoleTag' block-comment)?

lib/auth/role-manager.js
```diff
@@ -0,0 +1,463 @@
...
+            -1) {
+            isGlobalReader = true;
+        }
+    }
+
+    /*
+     * Check if we need to do per-request req.caller,
+     * either for role-operator or global read-only.
+     */
+    if (isRoleOper || isGlobalReader) {
```
danmcd, line 319: Are these clipped to far-less-than 80cols for a reason? Or was it copy/paste/indent that didn't in turn flesh things out to 80cols?  You may wanna fix that here.

danmcd (commented):
I like the smaller source files.  There appear to be some copy/paste/oops-didn't-push-out-to-80cols problems.

lib/aws-chunked-encoding.js
```diff
@@ -0,0 +1,128 @@
...
+ * AWS Chunked Encoding Handler for S3 Multipart Uploads
+ *
+ * For aws-chunked MPU, ensure metadata records use the decoded size
+ * per S3 spec: Content-Length is encoded size,
+ * x-amz-decoded-content-length is the actual payload size.
+ *
+ * Reference: AWS S3 API Documentation - Chunked Upload Encoding
+ * https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html
+ *
+ * CRITICAL: Not setting the decoded size correctly will cause Mako to fail
```
danmcd: Is this necessary when it's spelled out in more detail below?

lib/buckets/common.js
```diff
@@ -156,29 +163,146 @@ function _list(type, req, bucket_id) {
...
+        // Path traversal check
+        if (PATH_TRAVERSAL_REGEX.test(prefix)) {
+            process.nextTick(function () {
+                ee.emit('error', new InvalidParameterError('prefix',
+                    'Invalid prefix format'));
+            });
+            return (ee);
+        }
+
+        // Length check (AWS S3 has a practical limit around 1000 chars)
```
danmcd, line 204: Does this kneecap non-S3 uses ?

lib/buckets/delete.js
```diff
@@ -89,7 +89,6 @@ function deleteBucket(req, res, next) {
             var notFoundErr =
                 new errors.BucketNotEmptyError(bucket.name);
             bucketEmpty = false;
-            res.send(409, notFoundErr);
```
danmcd, line 92: I take it the return of `next(notFoundErr)` will percolate and an error gets sent elsewhere instead?

danmcd (commented):
INTENTIONAL checkpoint: resume with lib/buckets/cors.js.

Also a question:  With this refactor does the regression testing footprint go up or were regression tests already part of the plan?

danmcd:
> Also a question: With this refactor does the regression testing footprint go up or were regression tests already part of the plan?

Ooops, buckets haven't been used so there's not much to regression-test.  Sorry for the superfluous question.

lib/anonymous-auth.js
```diff
@@ -0,0 +1,630 @@
...
+    'authenticated-reader',
+    'owner-reader',
+    'owner-full-control',
+    'log-writer'
+]);
+
+///--- Configuration
+
+var ANONYMOUS_ACCESS_CONFIG = {
+    enabled: process.env.MANTA_ANONYMOUS_ACCESS_ENABLED === 'true' || true,
```
travispaul: Should this be:  `process.env.MANTA_ANONYMOUS_ACCESS_ENABLED === 'true' || false` 

As its written it seems this can never be disabled.

lib/anonymous-auth.js
```diff
@@ -0,0 +1,630 @@
...
+
+        callback(null, req.bucket.roles);
+        return;
+    }
+
+    // if the role is not found or is not public-read,
+    // just check if we need to allow access to the bucket/object
+    // based in naming.
+    // Only allow exact bucket name matches, not substring matches
+    var isPublicBucket = bucketName === 'public';
```
travispaul: Should `ANONYMOUS_ACCESS_CONFIG.enabled` also be consulted here?

lib/anonymous-auth.js
travispaul: I also don't see `allowedBuckets` consulted anywhere, is that still used?

lib/buckets/objects/copy.js
```diff
@@ -0,0 +1,921 @@
...
+ * - https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html
+ *
+ */
+
+/**
+ * Server-side copy implementation for S3 copy operations
+ * This function handles copying objects within Manta storage
+ * without requiring client-side download/upload.
+ */
+function copyObject(req, res, next) {
```
danmcd, line 52: Not now, but can this functionality be exported to non-S3 Manta file usage?

lib/buckets/objects/copy.js
```diff
@@ -0,0 +1,921 @@
...
+
+            callback(null, obj);
+        });
+}
+
+/**
+ * Perform the actual server-side copy operation
+ * Allocates destination sharks and streams data directly
+ */
+function performServerSideCopy(req, res, sourceReq, sourceMetadata, callback) {
```
danmcd, line 263: Per my comment above, maybe this function can be reachable by traditional manta someday?

lib/buckets/objects/copy.js
```diff
@@ -0,0 +1,921 @@
...
+ * - Stream object data to disk (sharks/mako)
+ *
+ * These are the transcendental operations which are the building blocks for
+ * every other operation.
+ *
+ * - https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html
+ *
+ */
+
+/**
```
danmcd, line 47: I can't remember if I asked this before or not, so apologies if I repeat myself.  If server-side copy fails partway through, will the successful part be marked as empty or garbage so it's not lying around eating space?

lib/aws-chunked-decoder.js
```diff
@@ -0,0 +1,477 @@
...
+ * Format:
+ * [chunk-size in hex];chunk-signature=[signature]\r\n
+ * [chunk-data]\r\n
+ * [chunk-size in hex];chunk-signature=[signature]\r\n
+ * [chunk-data]\r\n
+ * 0;chunk-signature=[final-signature]\r\n
+ * [optional trailing headers]\r\n
+ * \r\n
+ *
+ * This decoder strips the AWS chunk metadata and extracts only the actual data.
```
travispaul, line 141: I'm not seeing where we do the chunk signature verification and this comment hints that we're explicitly skipping that step. Is there a reason we're able to skip the signature chunk verification?

lib/buckets_mdapi_client.js
```diff
@@ -50,12 +51,67 @@ function createClient(options, callback) {
         }
         buckets_mdapi_args.unwrapErrors = true;
         buckets_mdapi_args.srvDomain = pnodeUrl.hostname;
-        buckets_mdapi_args.cueballOptions.defaultPort =
-            parseInt(pnodeUrl.port, 10);
-        buckets_mdapi_args.log = options.log.child({
+        buckets_mdapi_args.cueballOptions.defaultPort = parseInt(
+            pnodeUrl.port, 10);
+        // Create a completely filtered logger for buckets-mdapi to
+        // suppress user errors
```
danmcd, line 56: I might've asked this earlier... why suppress user errors?

lib/server/request-pipeline.js
```diff
@@ -0,0 +1,231 @@
...
+ * @param {Object} req - Restify request object
+ * @param {Object} res - Restify response object
+ * @param {Function} handleS3Request - S3 request routing function
+ * @param {Function} next - Restify next callback
+ */
+function processS3Request(req, res, handleS3Request, next) {
+    req.log.debug('S3_DEBUG: Processing S3 request through pipeline');
+
+    // Convert S3 AWS headers to ones that Manta could use/understand
+    s3Compat.s3HeaderTranslator(req, res, function (headerErr) {
```
travispaul, line 46: Nit (non blocker): Using vasync`s waterfall or pipeline function might help flatten the multiple nested callbacks here. Though admittedly, this function is simple enough that its not hard to follow so will leave this to your discretion.

lib/s3-routes.js
```diff
@@ -0,0 +1,2356 @@
...
+ */
+
+function s3DeleteBucketObjectsHandler() {
+    return function s3DeleteBucketObjects(req, res, next) {
+        req.log.debug('S3_DEBUG: s3DeleteBucketObjects' +
+            ' - bulk delete operation');
+
+        // Validate S3 request format
+        if (!req.s3Request || !req.s3Request.bucket) {
+            var InvalidBucketNameError = require('./errors').
```
travispaul: Can the `require()` calls in this function be hoisted to the top of the module?

lib/libmantalite.js
```diff
@@ -134,6 +134,8 @@ function authorize(opts) {
 
     var error;
 
+    // For assumed roles, IAM policy enforcement should be handled in auth.js
```
danmcd: Editorial:  "should be handled" ==> "should have been handled"

lib/cors-middleware.js
```diff
@@ -0,0 +1,593 @@
...
+
+    // Validate method
+    if (!isMethodAllowed(requestMethod,
+                         corsHeaders['access-control-allow-methods'])) {
+        log.debug('CORS: Method not allowed in CORS configuration');
+        res.send(403, 'CORS policy violation: Method not allowed');
+        return (next(false));
+    }
+
+    // Set CORS response headers for prefligh
```
danmcd: spelling: "preflight".  Also, did you mean to type more here?

lib/cors-middleware.js
```diff
@@ -0,0 +1,593 @@
...
+ * @note Supports both '*' and 'star' as wildcard values for backward
+ *       compatibility
+ * @note Origins are compared as exact strings (case-sensitive)
+ * @see processCorsOptions, convertBucketCorsToHeaders
+ *
+ * @example
+ * isOriginAllowed("https://example.com",
+ *                 "https://example.com,https://test.com") // true
+ * isOriginAllowed("https://evil.com", "https://example.com") // false
+ * isOriginAllowed("https://any.com", "*") // true
```
danmcd: Is `any.com` in this example a single entity? If so, `any.com` is a bad example.  Maybe `example2.com` or `another.com` instead?

danmcd (commented):
Checkpoint, resuming on iam-client.js

lib/iam-utils.js
```diff
@@ -0,0 +1,328 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Copyright 2025 Edgecast Cloud LLC.
+ */
+
+/*
+ * Common utility functions for IAM and STS request handling
+ * Eliminates code duplication across IAM/STS handlers
```
danmcd: Feels WAY too much like LLM... you can lose the "Eliminates code duplication..." sentence.

danmcd (commented):
Checkpoint after passing over the IAM code.

lib/metadata_placement.js
```diff
@@ -126,12 +127,55 @@ function MetadataPlacement(opts) {
 
             // Establish a buckets-mdapi connection to each shard
             self.log.info('creating buckets-mdapi clients');
+
+            // Create a filtered logger for buckets-mdapi to
+            // reduce user error noise
```
danmcd, line 126: I asked a similar question above:  Why reducing the error noise? There's better text here (I hope I didn't miss that better text above?), that suggests we don't want S3 clients denying-service on the server side with excessive messages. Is that the reason we suppress?

lib/policy-converter.js
```diff
@@ -0,0 +1,374 @@
...
+        allRules = ['DENY * *'];
+    }
+
+    // Remove duplicate rules
+    allRules = allRules.filter(function (rule, index, self) {
+        return (self.indexOf(rule) === index);
+    });
+
+    // Generate a UUID for the policy
+    var policyId = require('crypto').randomBytes(16).toString('hex');
```
danmcd: Do we generate proper-dashed UUIDs with this? Or does Manta Policy accept them as a single 32-digit hex string?

I don't see the use if "id" in this wad.  Am I missing it somewhere else?  I also don't seem to see sourceRole being used either apart from these return values.  Where do we USE these fields in manta policy? Because I'm not finding it!

danmcd (commented):
After looking at policy-converter.js, I'm seeing fields that are set but never used. I may be missing something, but github search and these *.patch files in the four-under-review show nothing but assignment/creation, but no use. See comments in policy-converter.js for examples.

Continue with s3-compat.js (but on a limited basis, so pardon lack of speed).

lib/metadata_placement.js
cneira: That's one reason: a 404 error when a client tries to reference a bucket or object that does not exist, or a combination of both, is a fairly common error that should not clutter the logs. There may be a better way to do it.

lib/buckets/objects/copy.js
cneira: Server Side copy does use multipart uploads, is just a regular PUT object operation, but is capped at 5GiB  object size. I have not checked on the storage node what happens if you interrupt an ongoing put object request. I'll need to check on that.

lib/buckets/objects/copy.js
cneira: It could, but would need more work; object operations in manta-buckets-api depend on buckets-mdapi and buckets-mdplacement for metadata, which are not used by traditional manta.

lib/buckets/objects/copy.js
cneira: No, without doing some legwork, we need to check how metadata for objects and streaming of data is done in traditional manta. As of today, manta-buckets-api have their own services for metadata.

lib/s3-routes.js
cneira: I'll scan all the files for this type of antipattern.

lib/server/request-pipeline.js
cneira: I'll take a look at vasync's waterfall if we could simplify it further.

lib/aws-chunked-decoder.js
cneira: The main reason was performance; to validate each chunk will require a trip to _mahi_ to validate the signature using Sigv4 per chunk, seems I did not keep track of the missing parts for aws-chunked. We accept aws-chunked transfer but we are not enforcing the security/integrity model that the aws chunked signature should impose. I think we  need to either remove aws-chunked transfers or implement the check verification. I'm more inclined now to remove aws-chunked transfers as this can be disabled with SDK clients https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/S3ClientOptions.Builder.html#disableChunkedEncoding--

Good catch!

lib/anonymous-auth.js
cneira: I'll revise the anonymous access configuration scheme, as it should always be allowed.

lib/buckets/delete.js
cneira: That will percolate and return an S3-compatible error to the client.

lib/buckets/common.js
cneira: Yes, that's right, but I think manta-buckets-api will only be used by S3 clients.

lib/anonymous-auth.js
cneira: In manta-buckets-api, anonymous access is granted to buckets called public, so if an HTTP request goes to
"https://manta.example.com/<username>/buckets/public/objects/<your object name>" should be able to grab the object without authentication. Maybe we could redirect /Joyent_Dev/buckets/public/objects/<object> to follow the same route that traditional Manta.
The anonymous access routes are part of https://github.com/TritonDataCenter/manta-buckets-api/blob/MANTA-5471-STS/docs/anonymous-access.md

lib/server/middleware.js
```diff
@@ -0,0 +1,507 @@
...
+
+/**
+ * Pre-middleware for comprehensive request logging and binary mode setup.
+ * Logs all request details for debugging and forces binary mode for S3 uploads.
+ *
+ * @param {Object} req - Restify request object
+ * @param {Object} res - Restify response object
+ * @param {Function} next - Restify next callback
+ */
+function logAllRequests(req, res, next) {
```
travispaul, line 213: Nit: I came here from  [lib/server.js](https://github.com/TritonDataCenter/manta-buckets-api/blob/506d44be608e4d4723dba09bcbca460f839d5f34/lib/server.js#L116C5-L116C43) and was surprised to find that this function was not only logging requests but also making modifications to the `req` object and enforcing binary mode for s3 requests.

That behavior is mentioned in the comment but it is unexpected from the function name alone. It might make sense to rename this function or split it into two separate concerns.

lib/server/middleware.js
```diff
@@ -0,0 +1,507 @@
...
+ * @param {Function} next - Restify next callback
+ */
+function preserveRawBodyPreMiddleware(req, res, next) {
+    // Check if this is a Complete Multipart Upload request
+    var isCompleteMultipartUpload = false;
+    if (req.method === 'POST' && req.url) {
+        // Simple check for uploadId query parameter
+        // (Complete Multipart Upload)
+        isCompleteMultipartUpload = req.url.includes('uploadId=') &&
+                                   !req.url.includes('uploads');
```
travispaul: Would this be incorrect if an object naturally has the string "uploads" in the path and is sent via mutlpart upload?

Also, I have to ask if we should we be using `req.query` here?

I see that this function is registered as a Restify [`pre()`](https://github.com/TritonDataCenter/manta-buckets-api/blob/506d44be608e4d4723dba09bcbca460f839d5f34/lib/server.js#L124) handler which happens before the query parser middleware is registered via  [`use()`](https://github.com/TritonDataCenter/manta-buckets-api/blob/506d44be608e4d4723dba09bcbca460f839d5f34/lib/server.js#L160) So you'd have to manually parse the params but that also makes me wonder if `preserveRawBodyPreMiddleware()` should be registered as an early `use()` middleware instead?

lib/server/middleware.js
```diff
@@ -0,0 +1,507 @@
...
+                      req.url === '/' &&
+                      req.headers.authorization &&
+                      req.headers.authorization.toLowerCase().
+        indexOf(constants.AWS_AUTH.SCHEME_SIGV4) !== -1;
+
+    if (((req.method === 'POST' &&
+          req.headers.authorization &&
+          (isXmlRequest || isCompleteMultipartUpload ||
+           isBulkDeleteRequest || isStsRequest)) ||
+         (req.method === 'PUT' && isCorsRequest)) &&
```
travispaul: `isCorsRequest` will only be `true` if `req.method === 'PUT'`

lib/server/middleware.js
```diff
@@ -0,0 +1,507 @@
...
+        req.headers['content-length'] &&
+        parseInt(req.headers['content-length'], 10) > 0) {
+
+        req.log.debug({
+            method: req.method,
```
travispaul: Just a suggestion if you weren't already aware. If you pass the whole `req` object to bunyan it will extract the [method, url, and headers](https://github.com/trentm/node-bunyan?tab=readme-ov-file#standard-serializers) for you.

If this is intentional and/or you were already aware feel free to disregard/resolve.  You might have good reasons for passing in explicit params too I just noticed the debug calls has a lot of properties. A custom serializer for common fields might be worth considering too.

lib/server/middleware.js
```diff
@@ -0,0 +1,507 @@
...
+ * Detects S3 requests by checking for SigV4 auth headers or presigned URLs,
+ * parses S3 request details, and marks binary operations.
+ *
+ * @param {Object} req - Restify request object
+ * @param {Object} res - Restify response object
+ * @param {Function} next - Restify next callback
+ */
+function s3RequestDetectorEarly(req, res, next) {
+    // Check if this is a SigV4 request and mark it for later processing
+    var authHeader = req.headers.authorization ||
```
travispaul: `req.header('Authorization', '')` is an equivalent drop-in replacement for this statement, which will fall back to an empty string if an authorization header (in either lower or capital case) is not provided.

The keys in `req.headers` are always lowercased by Node.js HTTP server so `req.headers.Authorization` will always be `undefined` here.

lib/anonymous-auth.js
```diff
@@ -0,0 +1,630 @@
...
+                uuidRoles: uuidRoles,
+                hasPublicReader: hasPublicReader
+            }, 'validateAnonymousObjectAccess:' +
+                          ' processed literal and UUID roles');
+            continueValidation();
+            return;
+        }
+
+        // Resolve remaining role UUIDs to names using Mahi
+        req.mahi.getName({
```
travispaul, line 597: `mahi.getName()` only appears to accept a `uuids` property: https://github.com/TritonDataCenter/node-mahi/blob/MANTA-5471/lib/client.js#L638

lib/anonymous-auth.js
```diff
@@ -0,0 +1,630 @@
...
+        path: req.path()
+    }, 'Marking request for potential anonymous access - will verify later');
+
+    // Extract the account name from the path (/account/buckets/bucket)
+    var pathParts = req.path().split('/').filter(function (part) {
+        return (part.length > 0);
+    });
+    var accountName = pathParts[0]; // First part is the account name
+
+    // Set up anonymous context tentatively - this will be validated later
```
travispaul, line 365: Do we actually validate this later? `accountName` get's passed around to a few places but mostly _seem_ to be only for logging purposes so we may not need to do anything but I didn't see any validation taking place (or I have missed it.)

travispaul (changes requested):
~~(Still reviewing but I've lost a comment possibly in an auto-collapsed file and hopes this resurfaces it)~~ Edit: Just reposting here:

The restify route handlers, and ideally any functions that are doing non-trivial work should be named (e.g. `function doSomething(){..}` instead of `function (){...}`) So that they're identifiable in stack traces.

lib/s3-compat.js
```diff
@@ -0,0 +1,2229 @@
...
+    'bucket-owner-read': ['owner-reader'],
+    'bucket-owner-full-control': ['owner-full-control'],
+    'log-delivery-write': ['log-writer']
+};
+
+
+
+/*
+ * Canned (XML format) Error Responses for S3 clients.
+ */
```
danmcd, line 65: These don't look like XML, why does the comment say they are?!

lib/s3-compat.js
```diff
@@ -0,0 +1,2229 @@
...
+ *  |    |                |                |           |   partNumber:1 }     |
+ *  +----+----------------+----------------+-----------+----------------------+
+ *  |GET | /mybucket\     | ListObjsV2     | path      | { bucket:"mybucket", |
+ *  |    | ?list-type=2   |                |           |   object:null,       |
+ *  |    |                |                |           |   operation:"ListV2",|
+ *  |    |                |                |           |   isS3Request:true,  |
+ *  |    |                |                |           |   mantaPath:null,    |
+ *  |    |                |                |           |   addrStyle:"path" } |
+ *  +----+----------------+----------------+-----------+----------------------+
+ *  |GET | bucket.domain\ | GetObj         | vhost     | { bucket:"bucket",   |
```
danmcd: Mention here briefly it's disabled, so when a reader continues they won't whiplash at the first entry in the Notes?

lib/s3-compat.js
```diff
@@ -0,0 +1,2229 @@
...
+ *   - x-amz-meta-*:      Any user defined metadata
+ *
+ * References:
+ *  - https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders\
+ *      .html
+ */
+function convertMantaToS3Headers(mantaHeaders, operation) {
+    var s3Headers = {};
+
+    Object.keys(mantaHeaders).forEach(function (key) {
```
danmcd, line 558: I saw @travispaul mentioning unnamed callback functions and here's another case of that.

lib/s3-compat.js
```diff
@@ -0,0 +1,2229 @@
...
+                    chunkLength: chunk.length,
+                    totalChunks: responseData.length
+                }, 'S3_DEBUG: s3ResponseFormatter - collected streaming chunk');
+            }
+            return (true); // Pretend write was successful
+        }
+        return (originalWrite.call(this, chunk));
+    };
+
+    res.end = function (endChunk) {
```
danmcd: Another one of @travispaul 's antipatterns.

lib/s3-mako-v2-commit.js
```diff
@@ -0,0 +1,528 @@
...
+
+/**
+ * Calculate object name hash for v2 storage path
+ */
+function calculateObjectNameHash(objectName) {
+    var crypto = require('crypto');
+    if (!objectName || typeof (objectName) !== 'string') {
+        throw new Error('objectName must be a non-empty string');
+    }
+    return (crypto.createHash('md5').update(objectName).digest('hex'));
```
danmcd: Couldn't you instead use the createMD5Hash you export in s3-compat?

lib/s3-multipart.js
```diff
@@ -0,0 +1,3235 @@
...
+ * - Handles instance crashes gracefully (lease expiration)
+ * - Scales across multiple manta-buckets-api instances
+ * - Provides observability through comprehensive logging
+ */
+function DistributedLockManager(req) {
+    this.req = req;
+    this.lockTimeout = 90000; // 90 seconds - sufficient for assembly + buffer
+    // 500ms between acquisition attempts for faster response
+    this.retryInterval = 500;
+    this.maxRetries = 150; // Maximum 30 seconds wait time (60 * 500ms)
```
danmcd: 60 * 500ms (30sec) or 150 * 500ms (75sec).  Comment doesn't match value.

danmcd (changes requested):
Continuing next with lib/s3-routes.js.

Noticed the @travispaul antipattern of no-named callbacks here too.

lib/policy-converter.js
cneira: > Do we generate proper-dashed UUIDs with this? Or does Manta Policy accept them as a single 32-digit hex string?

Mahi accepts them as a single 32-digit hex string https://github.com/TritonDataCenter/mahi/blob/7189f44922e9de5cf324b87e061096b2bfea6ef9/lib/server/redislib.js#L221
Just validates if the policyId is a string. But I agree we should use a UUID, as  has been done throughout the system

> I don't see the use if "id" in this wad. Am I missing it somewhere else? I also don't seem to see sourceRole being used either apart from these return values. Where do we USE these fields in manta policy? Because I'm not finding it!

The policyId is used when the mantaPolicy object is passed to Mahi PutRole endpoint  in  https://github.com/TritonDataCenter/manta-buckets-api/blob/506d44be608e4d4723dba09bcbca460f839d5f34/lib/iam-handlers.js#L540

lib/anonymous-auth.js
cneira: roles UUIDs should be already populated in https://github.com/TritonDataCenter/manta-buckets-api/blob/506d44be608e4d4723dba09bcbca460f839d5f34/lib/anonymous-auth.js#L466. 
From a previous `req.mahi.getUuid `call when getting object metadata https://github.com/TritonDataCenter/manta-buckets-api/blob/506d44be608e4d4723dba09bcbca460f839d5f34/lib/buckets/buckets.js#L296

travispaul: Sorry, what I was trying to say was that it seems the `account` and `role` properties aren't used or expected by `mahi.getName`, I noticed this while trying to trace where the user-supplied `accountName` variable was used.

lib/anonymous-auth.js
cneira: No, accountName is not being validated down the stack; we only rely on the role metadata from the object of interest in the request. This is a security issue; I'll add a lookup to at least verify if the account exists.

lib/s3-compat.js
cneira: The reason is that those errors are converted to XML by the `convertErrorToS3` function down the stack.

lib/server/middleware.js
cneira: Thank you! I did not know about that.

lib/server/middleware.js
cneira: Good catch, I have fixed this, along with the same error in CORS and bulk deletes

lib/s3-mako-v2-commit.js
cneira: done.

lib/s3-compat.js
cneira: All anonymous functions should have names by now.

lib/s3-multipart.js
cneira: fixed

lib/s3-compat.js
cneira: I just removed that record as we only support path-style access.

lib/s3-compat.js
cneira: Most anonymous functions that did complex work have already been named.

lib/aws-chunked-encoding.js
cneira: comment removed.

lib/buckets_mdapi_client.js
cneira: The reason is that those kinds of errors are pretty common, like a typo on the bucket name, so as to avoid cluttering the log these are skipped.

lib/libmantalite.js
cneira: This is fixed.

lib/anonymous-auth.js
cneira: Thank you!  I removed the non-required parameters.

lib/anonymous-auth.js
cneira: I just added a check to validate the account name.

lib/anonymous-auth.js
cneira: You are right, we are setting `isPublicBucket` flag, that only should happen in the context of anonymous authorization.

lib/anonymous-auth.js
cneira: I removed the not used property and allow anonymous access unless explicitly disabled by environment variables.

lib/anonymous-auth.js
```diff
@@ -0,0 +1,747 @@
...
+ *
+ * Supports multiple account name formats:
+ * - Human-readable: testuser, my-account, user_123
+ * - UUIDs: 550e8400-e29b-41d4-a716-446655440000
+ * - Mixed: user-123_prod.test
+ *
+ * @param {string} accountName - The account name to validate
+ * @returns {boolean} - True if valid format, false otherwise
+ *
+ * Validation rules:
```
travispaul, line 107: According to the [sdcperson schema](https://github.com/TritonDataCenter/sdc-ufds/blob/master/schema/sdcperson.js#L22) `@` is a legal character and a `-` is not allowed, account names must be between 3 and 32 chars and start with an alpha char.

I think we can copy the regexes from UFDS and do something similar to the following (doesn't have to be this code _exactly_ but the regexes should probably be taken verbatim):

```js

/**
 *  Regexes copied from ufds' schema/sdcperson.js
 *  It's a shame we don't have these in a re-usable shared module...
 *  Do we need at least the UUID_RE elsewhere in tihs repo too?
 */
var LOGIN_RE = /^[a-zA-Z][a-zA-Z0-9_\.@]+$/;
var UUID_RE = /^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$/;

/**
 * Validate account name format for security
 *
 * Account names extracted from URL paths must be validated to prevent
 * security vulnerabilities such as path traversal, injection attacks,
 * and authorization bypass.
 *
 * Supports multiple account name formats
 * - Account login: 3-32 chars, starts with letter, allowed chars are
 *   [A-Za-z0-9_.@]
 * - UUID: 36 chars, 8-4-4-4-12 hex
 * - Rejects path traversal characters/patterns
 *
 * @param {string} accountName - The account name to validate
 * @returns {boolean} - True if valid format, false otherwise
 *
 * Security invariant: All account names used in security contexts
 * MUST pass this validation.
 */
function isValidAccountName(accountName) {
    assert.string(accountName, 'accountName');
    
    // Nothing less than 3 chars can be valid
    if (accountName.length < 3) {
        return (false);
    }

    // Reject path traversal patterns explicitly
    if (accountName.indexOf('..') !== -1 ||
        accountName.indexOf('/') !== -1 ||
        accountName.indexOf('\\') !== -1) {
        return (false);
    }

    // Account names MUST be <= 32 characters
    if (accountName.length <= 32 && LOGIN_RE.test(accountName)) {
        return (true);
    }

    // ...else it MUST be a valid UUID
    return (accountName.length === 36 && UUID_RE.test(accountName));      
}

// sanity checks (does this cover all our bases?)
console.log(isValidAccountName('testuser') === true, 'Simple');
console.log(isValidAccountName('my-account') === false, 'Hyphen is not legal');
console.log(isValidAccountName('user_123') === true, 'Underscore');
console.log(isValidAccountName('user-123_prod.test') === false, 'Mixed - Hyphen not legal');
console.log(isValidAccountName('user123_prod.test') === true, 'Mixed');
console.log(isValidAccountName('t') === false, 'too short');
console.log(isValidAccountName('..') === false, 'too short');
console.log(isValidAccountName('/') === false, 'too short');
console.log(isValidAccountName('\\') === false, 'too short');
console.log(isValidAccountName('../') === false, 'path characters');
console.log(isValidAccountName('foo/') === false, 'path characters');
console.log(isValidAccountName('bar/') === false, 'path characters');
console.log(isValidAccountName('\\bar') === false, 'path characters');
console.log(isValidAccountName('foo\\') === false, 'path characters');
console.log(isValidAccountName('t@b') === true, '3 chars, @ is legal');
console.log(isValidAccountName('foo@bar') === true, '@ is legal');
console.log(isValidAccountName('@foobar') === false, 'must start with alpha');
console.log(isValidAccountName('foobar@') === true, '@ at end is legal');
console.log(isValidAccountName('a@@') === true, 'Oddly, legal can have multiple @');
console.log(isValidAccountName('0user') === false, 'must start with alpha');
console.log(isValidAccountName('550e8400-e29b-41d4-a716-446655440000') === true, 'lowercase uuid');
console.log(isValidAccountName('C8E03512-2BDF-4D8F-8785-372E092AD009') === false, 'not lowercase uuid' );
console.log(isValidAccountName('a01234567890123456789012345678901234567890123456789') === false, 'too long');

// will assert, passing in a non-string is a programming error more-so than user-input error in this case.
console.log(isValidAccountName(0) === false, 'not a string');
console.log(isValidAccountName(new Date()) === false, 'not a string');
console.log(isValidAccountName(true) === false, 'not a string');
console.log(isValidAccountName(null) === false, 'not a string');
console.log(isValidAccountName(undefined) === false, 'not a string');
```

travispaul: ~The fact `@` is permitted also makes me wonder if this function will encounter it as the URL encoded value of `%40` as part of a path...~
Update: I don't think we'll have to deal with any URL encoded `@`'s restify should return it as-is.

lib/s3-routes.js
travispaul: @cneira I cobbled together the command below to surface more examples of this, it has some false positives but spot checking a few shows some legit hits too:

```
$ git diff --name-only master -- '*.js' | xargs grep -n 'require(' | awk -F: '$2 >= 60 {print $2 "\t" $1 "\t" substr($0, index($0, $3))}' | sort -n
69	lib/buckets/options.js	            require('../common').CURRENT_STORAGE_LAYOUT_VERSION;
79	test/s3-operations.test.js	        var bucketServer = require('../lib/server');
81	lib/s3-mako-v2-commit.js	    var s3MultipartModule = require('./s3-multipart');
158	test/stack-trace-verification.test.js	            var module = require(modulePath);
172	lib/auth/role-manager.js	    var IAMClient = require('../iam-client');
203	lib/buckets/objects/copy.js	    var crypto = require('crypto');
234	lib/s3-mako-v2-commit.js	    var vasync = require('vasync');
237	lib/cors-middleware.js	        var corsNameHash = require('crypto').createHash('md5').
239	lib/buckets/cors.js	    var corsNameHash = require('crypto').createHash('md5')
242	lib/policy-converter.js	    var policyId = require('crypto').randomBytes(16).toString('hex');
250	lib/buckets/cors.js	        content_md5: require('crypto').createHash('md5')
264	lib/buckets/objects/copy.js	    var uuidv4 = require('uuid/v4');
299	lib/policy-converter.js	    var policyId = require('crypto').randomBytes(16).toString('hex');
317	lib/s3-multipart.js	var awsChunked = require('./aws-chunked-encoding');
333	lib/buckets/cors.js	    var corsNameHash = require('crypto').createHash('md5')
350	lib/auth/authorization-handler.js	            var iamPolicyEngine = require('../iam-policy-engine');
354	lib/aws-chunked-decoder.js	            var verifier = require('./aws-chunked-signature-verifier');
395	lib/buckets/objects/copy.js	    var sharkClient = require('../../shark_client');
396	lib/buckets/objects/copy.js	    var crypto = require('crypto');
416	lib/auth/role-manager.js	                    require('../errors').InvalidRoleError;
429	test/mock-infrastructure.js	var crypto = require('crypto');
430	test/mock-infrastructure.js	var stream = require('stream');
436	lib/buckets/cors.js	    var corsNameHash = require('crypto').createHash('md5')
499	lib/buckets/objects/copy.js	    var log = opts.log || require('bunyan').createLogger({name: 'copy'});
500	lib/buckets/objects/copy.js	    var crypto = require('crypto');
501	lib/buckets/objects/copy.js	    var vasync = require('vasync');
502	lib/buckets/objects/copy.js	    var sharkClient = require('../../shark_client');
604	lib/s3-multipart.js	    var createObjectModule = require('./buckets/objects/create');
868	lib/s3-routes.js	            var InvalidBucketNameError = require('./errors').
874	lib/s3-routes.js	        var InvalidParameterError = require('./errors').InvalidParameterError;
1018	lib/s3-routes.js	    var deleteObjectModule = require('./buckets/objects/delete');
1072	lib/common.js	                        require('./aws-chunked-signature-verifier');
1100	lib/s3-multipart.js	        hostname: require('os').hostname()
1190	lib/s3-routes.js	        var headBucket = require('./buckets/head');
1470	lib/s3-routes.js	                    require('./errors').InvalidRoleTagError;
1750	lib/s3-routes.js	        var objectGet = require('./buckets/objects/get');
1805	lib/common.js	                            require('./shark_client').storagePath({
1947	lib/s3-routes.js	        var copyObjectModule = require('./buckets/objects/copy');
2449	lib/s3-multipart.js	        var v2Multipart = require('./s3-mako-v2-commit');

```

cneira: Thank you! I pushed  some changes that should take care of this.

docs/CHG-006-NAMING-PATTERNS.md
```diff
@@ -0,0 +1,184 @@
+# CHG-006: Named Function Patterns for Stack Trace Improvements
```
danmcd: For those of us unversed, what *EXACTLY* is the origin of the term "CHG-006"?  You explain WHAT it is, but I don't know from where it originates.

lib/anonymous-auth.js
danmcd: Resolved

lib/aws-chunked-encoding.js
```diff
@@ -18,9 +18,6 @@
  * Reference: AWS S3 API Documentation - Chunked Upload Encoding
  * https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html
  *
```
danmcd, line 20: I take it this CRITICAL problem has gone away?

lib/s3-compat.js
danmcd: Resolving.

README.md
danmcd: This is still unaddressed.

lib/anonymous-auth.js
danmcd: Resolving.

lib/auth/authorization-handler.js
danmcd: Still unaddressed.

lib/auth/role-manager.js
danmcd: Still unaddressed or unanswered.

lib/aws-chunked-encoding.js
danmcd: Ooops, this was fixed.  Resolving (and removing another related).

lib/buckets/common.js
danmcd: Resolving.

lib/buckets/delete.js
danmcd: Resolving.

lib/buckets/objects/copy.js
danmcd: Thank you. Resolving.

lib/buckets/objects/copy.js
danmcd: Thank you, resolving.

lib/buckets/objects/copy.js
danmcd: Please let me know if this was tested and/or if there was a test-case added.

lib/buckets_mdapi_client.js
danmcd: Resolved, though bonus point if a comment here about it mentions the sheer amount of log-clutter.

lib/libmantalite.js
danmcd: Resolving.

lib/cors-middleware.js
danmcd: Unanswered.

lib/cors-middleware.js
danmcd: Still not fixed.

lib/cors-middleware.js
danmcd: Please review this and respond.

lib/iam-utils.js
danmcd: Resolving.

lib/metadata_placement.js
danmcd: Resolving, but like the other one, see if you can drop a bit more about log-noise and why into the comment.

lib/policy-converter.js
danmcd: Resolving.

danmcd (commented):
Checkpoint over incremental update, and older-comments I mentioned. Resolved older ones that got addressed (either with a good answer, or a code-change).

docs/CHG-006-NAMING-PATTERNS.md
travispaul: I was wondering about `NAMED-FUNCTIONS-REFACTORING.md` too. These look like LLM related instructions that accidentally got committed perhaps?

lib/aws-chunked-encoding.js
cneira: Yes, that's right, now we are validating the signature in aws-chunked-encoding.

docs/CHG-006-NAMING-PATTERNS.md
cneira: Yes, that's right, CHG-006.md and NAMED-FUNCTIONS-REFACTORING.md were LLM artifacts that should be removed.

lib/buckets/objects/copy.js
cneira: There is an integration test on test/run-s3-tests.sh  script for this  https://github.com/TritonDataCenter/manta-buckets-api/blob/6dd8e933d33e8c2f9645d47cbf1cc099cdd71626/test/test-scripts/s3-compat-awscli-test.sh.deprecated#L1344

lib/cors-middleware.js
cneira: Fixed

lib/cors-middleware.js
cneira: I updated the examples in this section to remove any.com

README.md
danmcd: It is now addressed.  Resolving.

lib/auth/authorization-handler.js
cneira: I added a comment that cross-account is not supported in manta-buckets-api and that's the reason we should not fall into this case statement.

danmcd: Resolving.

lib/aws-chunked-decoder.js
cneira: I had added validation for signature verification for chunks. As we already have access to the secret on the first Mahi call.

lib/buckets/objects/copy.js
danmcd: Big test file got refactored, looks like. I'll assume it landed somewhere and resolve this.

lib/cors-middleware.js
danmcd: While "another.com" isn't ideal, it much better and less confusing so I'm resolving this.

docs/CHG-006-NAMING-PATTERNS.md
danmcd: Resolving.

danmcd (commented):
Incremental.

Excellent, I hope I marked-as-resolved everything this wad fixed.

lib/cors-middleware.js
danmcd: Resolving

danmcd (commented):
Resolving a resolved comment.

lib/auth/role-manager.js
cneira: I'll fix that. I think I did `fold -w72 ` but the result was not good enough, and I forgot to update.

cneira: done

danmcd: Resolving.

danmcd (commented):
Incremental, thank you!

lib/auth/credential-provider.js
```diff
@@ -0,0 +1,669 @@
...
+require('../errors');
+
+
+///--- Functions
+
+function parseKeyId(req, res, next) {
+    // Skip parsing for S3 presigned URLs - they use different key format
+    if (req._s3PresignedAuthComplete) {
+        req.log.debug('parseKeyId: skipping' +
+                      ' - S3 presigned URL uses different key format');
```
danmcd: Grammar nit:  "uses a different key format".

danmcd (commented):
Only one nit so far.  Resuming with signature-verifier.js later tonight or first thing tomorrow.

lib/auth/signature-verifier.js
```diff
@@ -0,0 +1,1472 @@
...
+            // Check for X-Amz-Expires
+            if (isNaN(amzExpires) || amzExpires <= 0 || amzExpires > 604800) {
+                next(new PreSignedRequestError(
+                    'X-Amz-Expires must be a positive integer' +
+                    ' (max 604800 seconds/7 days)'));
+                return;
+            }
+
+            // Parse ISO8601 date: 20230101T120000Z (UTC)
+            var year = parseInt(amzDate.substr(0, 4), 10);
```
danmcd, line 168: Non-blocking nit:  This has a Y10K problem in it, but the solution is a detectable one (leading + or -) and not needed for 9000 years. Will resolve next time around.

lib/auth/signature-verifier.js
```diff
@@ -0,0 +1,1472 @@
...
+            var month = parseInt(amzDate.substr(4, 2), 10) - 1;
```
danmcd, line 169: Non-blocking TIL:  JS's Date() has 0-11 for months. Will resolve next time around.

lib/auth/signature-verifier.js
```diff
@@ -0,0 +1,1472 @@
...
+    // Skip signature verification for S3 presigned URLs
+    if (req._s3PresignedAuthComplete) {
+        req.log.debug('verifySignature: skipping ' +
+                      '- S3 presigned URL already validated');
+        next();
+        return;
+    }
+
+    // Skip signature verification for anonymous access
+    if (req.isAnonymousAccess) {
```
danmcd, line 1404: I think I asked this before: By the time we reach here we KNOW it's a legitimate anonymous access already, right?  IOW, `req` has been vetted and bad paths can't get us a req that lies about this, right?

lib/auth/token-manager.js
```diff
@@ -0,0 +1,318 @@
...
+ * Copyright 2019 Joyent, Inc.
+ * Copyright 2026 Edgecast Cloud LLC.
+ */
+
+/*
+ * token-manager.js: Authentication token creation and
+ * parsing.
+ *
+ * Handles encryption/decryption of authentication tokens
+ * using AES-128-CBC with gzip compression.
```
danmcd, line 18: Is AES-128-CBC an S3 standard?  IF SO, treat the rest of this as a non-blocking scold of the spec.

I ask because AES-128-CBC is an unauthenticated block cipher, and it's only 128-bits vs. 192 or 256 bits.  Although if the encrypted token is what we're using for authentication, I might be over blowing. (Though as a rule encrypt-then-auth outbound and auth-then-decrypt inbound is the best practice.)

If it's something else, we should make this more agile, unless this lock-in requires flag-day level intervention.

Sorry if I didn't catch this sooner (or if I asked previously).

lib/aws-chunked-signature-verifier.js
```diff
@@ -0,0 +1,161 @@
...
+ */
+
+/*
+ * Copyright 2026 Edgecast Cloud LLC.
+ */
+
+/*
+ * AWS SigV4 Chunked Signature Verification
+ *
+ * This module provides cryptographic utilities for verifying AWS SigV4
```
danmcd, line 14: Non-blocking will-resolve next time:  This feels like AWS/S3 standards. Am I right? I like the use of HMAC-SHA256 (HMAC is still a wonderful construction even 30+ years later IMHO).

danmcd (commented):
Checkpoint 2, resuming with cors.js.  As with mahi, may not be reviewing tests (may browse).

danmcd (commented):
No-comment checkpoint.  Resume with lib/common.js.

lib/common.js
```diff
@@ -154,51 +154,427 @@ HttpRequest.isBucketObject = function isBucketObject() {
...
+
+    if (md && Object.keys(md).length > 0) {
+        var mdKeys = Object.keys(md);
+        req.log.debug({
+            metadataKeysCount: mdKeys.length,
+            firstFewKeys: mdKeys.slice(0, 10)
+        }, 'CORS DEBUG: Processing metadata headers');
+
+        // Log each metadata key individually to see exact names
+        mdKeys.forEach(function (k) {
```
danmcd, line 184: Unnamed functions? But original(?) code had it too so maybe not to worry here?

And sorry if I missed this in other places previous to this.

danmcd (commented):
One nit about callback function names, @travispaul may have opinions on that as well.

Checkpoint, resuming with cors-middleware.js, and did some work from the bottom-up so last file left unreviewed is sts-handlers.js.

lib/sts-handlers.js
```diff
@@ -0,0 +1,390 @@
...
+
+        log.debug({
+            query: req.query,
+            body: req.body,
+            headers: req.headers
+        }, 'STS AssumeRole request received');
+
+        // Parse STS parameters (can come from query or body)
+        // awscli sends them in body.
+        var roleArn = req.query.RoleArn || req.body.RoleArn;
```
danmcd: As it stands now: if for some bizarre reason RoleArn is in BOTH places, the query's RoleArn takes precedence over the body's RoleArn. This is desirable or standard-specified? Or should we swap the places to body-first then query-first?

There are other query-first `||` patterns below like this one that pick query over body. IF I see an inconsistency I'll highlight it. Also if body needs to take precedence there will be a lot of swapping. :(

danmcd: We discussed this on Slack, I know.  One of us will make sure your followup with the spec is mentioned here.

lib/policy-converter.js
```diff
@@ -0,0 +1,374 @@
...
+var crypto = require('crypto');
+
+/**
+ * AWS S3 Action to Manta Action mapping
+ */
+var AWS_TO_MANTA_ACTIONS = {
+    // Bucket operations
+    's3:ListBucket': 'getbucket',
+    's3:ListBuckets': 'getbucket',  // Special case for listing all buckets
+    's3:GetBucket': 'getbucket',
```
danmcd, line 28: Documented in `Authorization Action Mapping` of docs/architecture.md.  Will resolve next time around.

lib/s3-mako-v2-commit.js
```diff
@@ -0,0 +1,533 @@
...
+/*
+ * Copyright 2026 Edgecast Cloud LLC.
+ */
+
+/*
+ * File:     s3-mako-v2-commit.js
+ * Purpose:  Extends s3-multipart.js with /mpu/v2/commit endpoint support
+ *
+ * This method relies in mako endpoint /mpu/v2/commit endpoint to be available.
+ * Assembling all the parts using mako, simplifies the code and the responsabil-
```
danmcd: spelling nit:  "respons*i*bility"

danmcd (commented):
Last checkpoint before completion.  Only the very large s3-multipart.js and s3-routes.js remain.

lib/auth/credential-provider.js
cneira: Fixed grammar.

lib/s3-multipart.js
```diff
@@ -0,0 +1,3254 @@
...
+ */
+
+///--- S3 Multipart Upload Handlers
+
+/**
+ * S3 Initiate Multipart Upload Handler
+ * POST /{bucket}/{key}?uploads
+ */
+function s3InitiateMultipartUploadHandler() {
+    return function s3InitiateMultipartUpload(req, res, next) {
```
danmcd, line 81: Okay, I might be missing something subtle, but this handler() that has a WHOLE FUNCTION INSIDE OF IT feels... off. Why not just have s3InitiateMultipartUpload() be the exported function?

Is there some required indirection I'm missing?  Feels like some JS idiom I'm not understanding. It almost feels like a JS construction to provide a function pointer of some sort.

lib/s3-multipart.js
```diff
@@ -0,0 +1,3254 @@
...
+    executePartUpload(req, res, next, partReq, etagCapture, partNumber,
+        partId, partKey, uploadId);
+}
+
+/**
+ * S3 Upload Part Handler
+ * PUT /{bucket}/{key}?partNumber={partNumber}&uploadId={uploadId}
+ */
+function s3UploadPartHandler() {
+    return function s3UploadPart(req, res, next) {
```
danmcd, line 647: Same oddness here.  Did I ask this question before and get an answer from you that I can't remember?

lib/s3-multipart.js
```diff
@@ -0,0 +1,3254 @@
...
+        });
+    } else {
+        applyCorsAndSendXML(req, res, next, xml);
+    }
+}
+
+/**
+ * Create distributed lock manager instance
+ */
+function createLockManager(req) {
```
danmcd, line 821: Why not just inline this in its caller?

lib/s3-multipart.js
```diff
@@ -0,0 +1,3254 @@
...
+}
+
+/**
+ * Execute completion workflow with distributed locking
+ * Acquires lock, calculates size, creates commit body, and executes assembly
+ */
+function executeCompletionWithLock(req, res, next, uploadRecord, partsFromXML,
+    owner, uploadId) {
+
+    var lockManager = createLockManager(req);
```
danmcd, line 925: Per earlier comment, why not just `lockManger = new DistributedLockManager(req);` here?

danmcd (commented):
Checkpoint mid-file: continue on line 973 of s3-multipart.js.

lib/s3-multipart.js
```diff
@@ -0,0 +1,3254 @@
...
+
+/**
+ * Retry wrapper for lock acquisition
+ * Handles retry logic with max attempts and interval
+ */
+function retryLockAcquisition(self, lockParams, maxRetries, retryInterval,
+    callback) {
+    var currentRetry = 0;
+
+    function attemptWithRetry() {
```
danmcd, line 1465: Okay... I was wondering WTF about function-in-the-function HERE, but please confirm that "currentRetry" outside the scope here, and function-as-parameter inside itself, is why it's constructed this way?

Either @cneira or @travispaul can answer this one, it's my lack of JS knowledge that's hurting me here, I think.

lib/auth/token-manager.js
cneira: The AES-128-CBC was used in the original manta-buckets-api https://github.com/TritonDataCenter/manta-buckets-api/blob/cc3e5d712891e7ab6f147e0eca9d2fa238ed6958/lib/auth.js#L36 is used by the legacy manta-buckets-api but no for S3 related features.

lib/s3-multipart.js
```diff
@@ -0,0 +1,3254 @@
...
+                        'NO-VALUE',
+                    getError: getErr ? getErr.message : 'no-error'
+                }, 'S3_MPU: Raw buckets-mdapi result debug');
+
+                if (getErr) {
+                    return (callback(getErr));
+                }
+
+                try {
+                    req.log.debug({
```
danmcd, line 2064: Is there anything in this try block that throws something to catch below?  Doesn't look like it?  Unless it's just the callback at the end of the block, in which case why is the try wrapped around all of the leadup to it?

lib/s3-multipart.js
```diff
@@ -0,0 +1,3254 @@
...
+    });
+
+    xml += '</ListPartsResult>\n';
+    return xml;
+}
+/* END JSSTYLED */
+/**
+ * Handle resume upload requests (PUT without partNumber)
+ */
+function resumeUploadHandler() {
```
danmcd, line 3144: Okay, this smaller example is easier to encapsulate... I see resumeUploadHandler being used in s3-routes.js so I'm guessing this gets `req, res, next` from some framework, right?

danmcd (commented):
Checkpoint:  Only s3-routes.js left to do.  After lunch.

lib/s3-multipart.js
cneira: Here we are using a factory pattern as it makes it simpler to mock-up createLockManager in tests, also allows us to have only one point of change if we need to add a new locking mechanism, the change would be extending createLockManager to add one, without needing to modify the call sites.

lib/s3-multipart.js
cneira: Using this function factory pattern allows us to extend this easily in the future if we need to.

lib/s3-multipart.js
cneira: The factory function pattern here is just for allow extensibility if we need to modify this endpoint in the future.

lib/s3-multipart.js
cneira: If we use the factory function pattern here, it will allow us to extend this in the future with less work.

lib/s3-mako-v2-commit.js
cneira: Fixed

lib/aws-chunked-signature-verifier.js
cneira: Yes, that's what AWS/S3 uses.

lib/common.js
cneira: When we iterate objects I think is ok to use lambdas, when we are doing more error prone/complex things having a named function on the stack trace will help a lot.

lib/s3-mako-v2-commit.js
danmcd: Resolving.

lib/s3-routes.js
```diff
@@ -0,0 +1,2372 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Copyright 2026 Edgecast Cloud LLC.
```
danmcd, line 8: NIT: Copyright in own block.

lib/s3-routes.js
```diff
@@ -0,0 +1,2372 @@
...
+            next(err);
+            return;
+        }
+        if (currentIndex >= flatChain.length) {
+            req.log.debug('S3_DEBUG: executeMiddlewareChain -' +
+            'completed all handlers, calling completion callback');
+            next();
+            return;
+        }
+        var currentHandler = flatChain[currentIndex];
```
danmcd, line 121: Right here...

lib/s3-routes.js
```diff
@@ -0,0 +1,2372 @@
...
+        req.log.debug({
+            handlerIndex: currentIndex,
+            handlerName: currentHandler.name || 'anonymous'
+        }, 'S3_DEBUG: executeMiddlewareChain - executing handler');
+
+        currentIndex++;
+        currentHandler(req, res, runNext);
```
danmcd, line 128: ... and right here is what I *think* is the reason why we have these weird functions-that-return-functions in the other file.  We have these systems of execution, and we want to squirrel-away the parms until we actually need to CALL the returned-function.

Still feels a little like too-much-indirection, but with this I think we don't have to worry right now about changing a bunch of code.  'dunno if this is a typical Javascript idiom that I'm just not familiar with, or if it needs an explanation somewhere, but that's for the more JS-savvy in the audience to answer.

lib/s3-routes.js
```diff
@@ -0,0 +1,2372 @@
...
+
+        // Convert S3 parameters to Manta format
+        req.params.account = req.caller.account.login;
+        req.params.bucket_name = bucketName;
+        req.params.object_name = objectPath;
+
+        // Call the original Manta delete object handler
+        var mantaHandlerChain = buckets.deleteBucketObjectHandler();
+        executeMiddlewareChain(mantaHandlerChain, req, res,
+                               function onDeleteBucketObjectComplete(error) {
```
danmcd, line 855: Also, probably missed this earlier, but with chain-in-chains and the need to flatten them to a single linear sequence, no wonder you need the indirection, right?!

lib/s3-routes.js
```diff
@@ -0,0 +1,2372 @@
...
+var headBucket = require('./buckets/head');
+
+///--- Helper Functions
+
+/**
+ * Execute a middleware chain from buckets module
+ * Flattens nested arrays that may be returned by handler factories
+ */
+function executeMiddlewareChain(chain, req, res, next) {
+    // Flatten the chain to handle nested arrays from handler factories
```
danmcd, line 61: I thought executeMiddlewareChain was common between s3-multipart.js and here, but clearly it's not currently.

Is this *deliberate* or was this a refactoring that needs to be cleaned up?  I can't tell which it is because THIS version of the function is larger than the one in s3-multipart.js.

Please supply a definite answer here, even if it's "should be refactored but the overhead is SO LARGE that we can't do it right now."  If it is so large, is a rename possible (say renaming s3-multipart.js's to `executeMPUMiddlewareChain`) or would that also cause more problems than it solves?

danmcd (commented):
ALMOST end of pass (hit submit by accident).

One nit, one big thing (that I will understand if it doesn't get addressed immediately), and a bunch of I-think-I-get-this-now.  Hopefully I'm not wrong in my newfound understanding.

danmcd (commented):
PASS COMPLETE.  Okay, so I guess I could've called it done with the last one.

Most of what confuses me at this point is MOST LIKELY attributable to my lack of JavaScript-fu.

lib/auth/signature-verifier.js
cneira: Yes, that's correct. The only part where this flag is set to true is on `validateAnonymousObjectAccess`

lib/s3-multipart.js
cneira: Having `attemptWithRetry()` inside retryLockAcquisition has access to currentRetry (closure), but mostly the reason is to be able to call `retryLockAcquisition`  async.

lib/s3-multipart.js
cneira: `JSON.parse` could throw [SyntaxError ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse)

lib/s3-routes.js
cneira: Fixed.

lib/s3-routes.js
cneira: We should refactor this to be common between modules; this is tech debt that I would prefer to pay on the next iteration.
The main reason is that this refactor will require more time, and I would like to focus on this stage on bug fixes as a priority, then tackle tech debt.

lib/s3-routes.js
cneira: You are absolutely correct on this. This is the intention to reduce the effort when we need to extend or modify code.

lib/s3-routes.js
cneira: Yes, that's correct, I needed to flatten the resulting object to process it.

lib/s3-multipart.js
cneira: When the route is registered, the `resumeUploadHandler` will return the inner function, so then when the route is hit, req, res, and next are provided by the Restify framework to `resumeUpload`

lib/auth/signature-verifier.js
cneira: We are following the Sigv4 spec where X-Amz-Date should be YYYYMMDD'T'HHMMSS'Z'  https://docs.aws.amazon.com/organizations/latest/APIReference/CommonParameters.html
I hope that somewhere along the line, a new specification addresses this.

lib/s3-routes.js
danmcd: Resolving.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF.  Thank you.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF.  Thank you.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF.  Thank you.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF.  Thank you.

lib/s3-multipart.js
danmcd: Thank you, resolving.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF.  Thank you.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF.  Thank you.

lib/sts-handlers.js
danmcd: Thank you. Once I see it I can resolve this conversation.

lib/common.js
danmcd: Understood, resolving.

lib/aws-chunked-signature-verifier.js
danmcd: Nice.  Resolving.

lib/auth/token-manager.js
danmcd: Ahhh, S3 uses the HMAC-SHA256 like I mentioned below.  Gotcha.  I would've dinged someone on this had I reviewed this back in the day.

lib/auth/signature-verifier.js
danmcd: Resolving.

lib/auth/signature-verifier.js
danmcd: If there is mention of ISO8601 in the spec you cite, they can take in the 8601 extension that has + and - prefixes.  Resolving.

lib/s3-routes.js
danmcd: Resolving now that I understand WTF. Thank you.

lib/s3-routes.js
danmcd: Resolving now that I understand WTF. Thank you.

lib/s3-routes.js
danmcd: Resolving now that I understand WTF. Thank you.

lib/s3-routes.js
danmcd: Resolving now that I understand WTF. Thank you.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF. Thank you.

lib/s3-multipart.js
danmcd: Resolving now that I understand WTF. Thank you.

danmcd (commented):
Only one last unresolved which needs a documentation/comment fix.

lib/sts-handlers.js
cneira: I just updated the code for STS and IAM handler to take body precedence over query in POST requests.

danmcd: THANK YOU, I'm glad my sense of these things was not out of order.

I looked over the precedence tests too as part of this. And they seem sane enough to resolve this.

danmcd (dismissed):
Thank you, and VERY glad you check the spec about body vs. request.

danmcd:
p.s. As with mahi, make sure you have package.json's in synch to prevent bad things from happening.

danmcd (approved):
Re-approving.  Non-blocking question, though:  This big wad of changes in the tests is to make sure we use library/production code whenever possible, and because of that we export more from `lib/` and re-implement less in `test/`, right?

cneira:
> Re-approving. Non-blocking question, though: This big wad of changes in the tests is to make sure we use library/production code whenever possible, and because of that we export more from `lib/` and re-implement less in `test/`, right?

Exactly, that was the point, previous tests were subject to be of no use if the helper functions changed in the actual production code, so now the code from lib/ is exercised.

travispaul (approved):

PR #71, opened 2026-03-06
MANTA-5507 manta-buckets-api should support rclone

cneira:
This PR includes the following tickets :

- MANTA-5507
- MANTA-5508
- MANTA-5509 
- MANTA-5510
- MANTA-5512
- MANTA-5513
- MANTA-5514

And fixes for problems I discovered while testing the checksum and sync features of rclone, as it uses the server-side copy that we expose, which surfaced some issues that I have addressed in this PR.

 Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/lib/auth/signature-verifier.js
+++ b/lib/auth/signature-verifier.js
@@ -1046,6 +1046,11 @@ function sigv4Handler(req, res, next) {
                 'accesskeys');
         }
 
+        if (result.signingKey) {
+            req.auth.signingKey = Buffer.from(
+                result.signingKey, 'hex');
+        }
+
         req.log.debug({
             accessKeyId: result.accessKeyId,
             userUuid: result.userUuid,
@@ -1155,6 +1160,11 @@ function sigv4Handler(req, res, next) {
                 }, 'Secret access key not in verifySigV4 response - ' +
                     'will be loaded by credential-provider from account ' +
                     'accesskeys');
+            }
+
+            if (result.signingKey) {
+                req.auth.signingKey = Buffer.from(
+                    result.signingKey, 'hex');
             }
 
             req.log.debug({
--- a/lib/auth/signature-verifier.js
+++ b/lib/auth/signature-verifier.js
@@ -801,240 +801,17 @@ function sigv4Handler(req, res, next) {
             filteredHeaders['content-md5'];
     }
 
-    var urlForVerification = req.url;
-    var pathPart = urlForVerification.split('?')[0];
-    var queryPart = urlForVerification.split('?')[1];
-
-    // For virtual-hosted-style requests, the bucket name is the first
-    // component of the host header.
-    var hostHeader = req.headers.host || '';
-    var virtualHostBucketName = hostHeader.split('.')[0];
-
-    // For path-style requests, extract bucket name from the path
-    var pathStyleBucketName = null;
-    var pathSegments = pathPart.split('/').filter(
-        function isNonEmptySegment(segment) {
-        return (segment.length > 0);
-    });
-    if (pathSegments.length >= 1) {
-        pathStyleBucketName = pathSegments[0];
-    }
-
-    // Get user agent for client-specific logic
-    var userAgent = req.headers['user-agent'] || 'none';
-
-    // Determine the actual bucket name based on request style
-    var bucketName = null;
-    if (virtualHostBucketName !== hostHeader &&
-        virtualHostBucketName.length > 0) {
-        // Virtual-hosted style: bucket.domain.com
-        bucketName = virtualHostBucketName;
-        req.log.debug({
-            bucket: bucketName,
-            style: 'virtual-hosted'
-        }, 'S3_AUTH_DEBUG: Detected virtual-hosted bucket name');
-    } else if (pathStyleBucketName) {
-        // Path-style: domain.com/bucket
-        bucketName = pathStyleBucketName;
-        req.log.debug({
-            bucket: bucketName,
-            style: 'path-style'
-        }, 'S3_AUTH_DEBUG: Detected path-style bucket name');
-    }
-
-    // Check if this is a presigned request
-    // (which uses proper path canonicalization)
-    var isPresignedRequest = queryPart &&
-        (queryPart.includes('X-Amz-Algorithm') ||
-         queryPart.includes('AWSAccessKeyId'));
-
-    // Detect client types and their trailing slash behavior
-    // CLI tools and SDK v2 sign WITHOUT trailing slash
-    // SDK v3 signs WITH trailing slash
-
-    // AWS SDK v3 specifically (aws-sdk-js/3.x) - signs WITH trailing slash
-    var isAwsSdkV3 = userAgent &&
-        (userAgent.toLowerCase().includes('aws-sdk-js/3') ||
-         (userAgent.toLowerCase().includes('aws-sdk-js') &&
-          userAgent.match(/aws-sdk-js\/3/i)));
-
-    // CLI tools and AWS SDK v2 (aws-sdk-nodejs) - sign WITHOUT trailing slash
-    var isAwsCli = (userAgent &&
-        (userAgent.toLowerCase().includes('aws-cli') ||
-         userAgent.toLowerCase().includes('boto3') ||
-         userAgent.toLowerCase().includes('botocore') ||
-         userAgent.toLowerCase().includes('aws-sdk-nodejs'))) ||
-        isPresignedRequest;
-
-    // DEBUG: Log the values we're working with
-    req.log.debug({
-        bucketName: bucketName,
-        pathPart: pathPart,
-        pathMatches: bucketName && pathPart === '/' + bucketName,
-        isAwsCli: isAwsCli,
-        isAwsSdkV3: isAwsSdkV3,
-        userAgent: userAgent
-    }, 'S3_AUTH_DEBUG: Pre-bucket path logic values');
-
     /*
-     * If the URI path is the bucket name, some S3 clients sign the request
-     * with a canonical URI ending with /. We must normalize the URL we send
-     * to mahi for verification to match what the client signed.
-     * AWS CLI/boto/SDK v2 do NOT add trailing slash
-     * AWS SDK v3 and other clients DO add trailing slash
+     * Use the original URL (_rawSignedUrl) for SigV4, before
+     * MANTA-331 sanitizePath() stripped trailing slashes.
      */
-    var bucketPathHandled = false;
-    if (bucketName && pathPart === '/' + bucketName) {
-        bucketPathHandled = true;
-        if (isAwsCli) {
-            // AWS CLI, boto*, and SDK v2 sign WITHOUT trailing slash
-            req.log.debug({
-                bucket: bucketName,
-                originalPath: pathPart,
-                userAgent: userAgent
-            }, 'S3_AUTH_DEBUG: CLI/SDK v2 detected -' +
-                          ' NOT adding trailing slash for bucket URI');
-            // pathPart stays as-is (no trailing slash)
-        } else {
-            // AWS SDK v3 and other S3 clients sign WITH trailing slash
-            req.log.debug({
-                bucket: bucketName,
-                originalPath: pathPart,
-                newPath: '/' + bucketName + '/',
-                userAgent: userAgent,
-                isAwsSdkV3: isAwsSdkV3
-            }, 'S3_AUTH_DEBUG: SDK v3/S3 client detected - ' +
-                'adding trailing slash for bucket URI SigV4 verification');
-            pathPart = '/' + bucketName + '/';
-        }
-    }
-
-    /*
-     * MANTA-331 removes trailing slashes, here minio requires them
-     * to sign the request if pathPart is just a bucket request then add /
-     * Most S3 clients need trailing slash appended to bucket paths for
-     * proper SigV4 signature validation. This is needed because Manta strips
-     * trailing slashes (manta 331) but clients calculate signatures with them.
-     * AWS CLI/boto/SDK v2 are excluded as they sign without trailing slashes.
-     * AWS SDK v3 and other clients sign WITH trailing slashes.
-     * For example /bucket needs to be /bucket/ for signature validation,
-     * while /bucket/file.txt doesn't need patching.
-     */
-    var regex = /\/[^/]+\.(?:$|\/)|\/[^/]+\/[^/]+/;
-    // Exclude directory creation requests - they are handled at URL
-    // reconstruction
-    var isDirectoryCreation = req.headers['content-type'] ===
-        'application/x-directory';
-
-    // Add slash for all S3 clients EXCEPT AWS CLI/boto/SDK v2
-    var shouldAddSlash = !bucketPathHandled &&
-        !isAwsCli &&
-        !regex.test(pathPart) &&
-        pathPart !== '/' &&
-        !pathPart.endsWith('/') &&
-        !isDirectoryCreation;
+    var urlForVerification = req._rawSignedUrl || req.url;
 
     req.log.debug({
-        userAgent: userAgent,
-        pathPart: pathPart,
-        regexTest: regex.test(pathPart),
-        shouldAddSlash: shouldAddSlash,
-        isDirectoryCreation: isDirectoryCreation,
-        bucketPathHandled: bucketPathHandled,
-        isAwsCli: isAwsCli,
-        isAwsSdkV3: isAwsSdkV3,
-        originalPath: pathPart
-    }, 'S3_AUTH_DEBUG: Trailing slash logic evaluation');
-
-    if (shouldAddSlash) {
-        req.log.debug({
-            originalPath: pathPart,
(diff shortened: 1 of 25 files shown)


danmcd (commented):
Which bug(s) do(es) this fix?

cneira:
> Which bug(s) do(es) this fix?

It fixes bug described in MANTA-5507, where rclone is not able to authenticate, and the rest of PRs add functionality required by rclone, which are bugs and features missing to support more of Server Side Copying features.

lib/auth/signature-verifier.js
```diff
@@ -104,6 +104,15 @@ function convertS3PresignedToManta(req, res, next) {
 
     // Check if this is an S3 presigned URL
     // (only convert if it has S3-specific parameters)
+    // Reject early if Algorithm/Credential are present but Signature
+    // is missing or empty — this is a malformed presigned URL that
+    // must not fall through to the normal (unauthenticated) path.
+    if (req.query['X-Amz-Algorithm'] && req.query['X-Amz-Credential'] &&
```
danmcd, line 110: Does this fix a *security* issue as well as the functional issue?  Or is this required functional (and security) changes for rclone?

IOW: Are there pre-existing *security* issues fixed in this PR?

lib/buckets/objects/get.js
```diff
@@ -82,7 +82,14 @@ function getObject(req, res, next) {
             common.CURRENT_STORAGE_LAYOUT_VERSION;
 
         // Add other needed response headers
-        res.set('Etag', object_data.id);
+        if (req.isS3Request) {
```
danmcd: Could you collapse this check (which I see a lot in this PR) into the base `set()` method itself?

lib/buckets/buckets.js
```diff
@@ -413,6 +413,25 @@ function conditionalHandler(req, res, next) {
...
+        var matched = false;
+
+        if_match.forEach(function (client_etag) {
+            if (client_etag === '*' || client_etag === etag) {
+                matched = true;
+            }
+        });
+
+        if (!matched) {
+            code = 412;
```
danmcd, line 434: Comment with context for code 412?

lib/common.js
```diff
@@ -67,6 +67,23 @@ var DEF_NUM_COPIES = constants.COPY_LIMITS.DEFAULT_COPIES;
 // The MD5 sum string for a zero-byte object.
 var ZERO_BYTE_MD5 = '1B2M2Y8AsgTpgAmY7PhCfg==';
 
+/*
+ * Convert base64-encoded MD5 to 32-char hex string
+ * for S3 ETag compatibility.  Returns null if input
+ * is falsy or not valid base64.
```
danmcd: "falsy"?

If it's a JS term of art (e.g. `!base64Md5`), the comment up here is a hair pedantic?

If it's not a JS term of art, then a more human-readable, "empty or invalid base64", may be more appropriate.

test/test-scripts/validate-manta-5512.sh
```diff
@@ -0,0 +1,327 @@
+#!/bin/bash
+# Copyright 2026 Edgecast Cloud LLC.
+# Validation script for MANTA-5512:
+# "Use mahi-derived signing key for AWS chunked STS verification"
+#
+# Verifies that aws-chunked uploads work with STS temporary credentials.
+# Before this fix, per-chunk signature verification failed because the
+# secret access key for STS session keys is not in the account's
```
danmcd, line 8: This paragraph makes me feel like my question about is-this-a-security fix is less pertinent, or outright extraneous.

danmcd (commented):
Checkpoint with the longer s3-multipart.js and s3-routes.js code changes and the longer mc-validation-test.sh and rclone-validation-test.sh scripts remaining.

lib/auth/signature-verifier.js
cneira: This fixes a functional issue (crash when signature is empty) which is also a security issue (not short circuiting when the signature was empty).
The other security issue regarding aws-chunked transfers is already addressed in MANTA-5512.

test/test-scripts/validate-manta-5512.sh
cneira: MANTA-5512 is a security fix, this only happens in access keys that are temporary (obtained through STS). In this case every  chunk of a aws-chunked transfer is not validated.

lib/auth/signature-verifier.js
danmcd: Thank you.  Resolving.

test/test-scripts/validate-manta-5512.sh
danmcd: Thank you for the clarification.  "Less pertinent," now means, "less pertinent because it's fixed in this wad." Resolving.

lib/s3-routes.js
```diff
@@ -277,6 +278,27 @@ function s3CreateBucketHandler() {
         req.params.account = req.caller.account.login;
         req.params.bucket_name = req.params.bucket;
 
+        // Most S3 clients just accept a 2XX HTTP response
```
danmcd, line 279: I caught a constant earlier, and the code-comments around here are the kind of code-comments I want to see around the other one.

danmcd (commented):
I think I have a few open comments still but first pass is done.

lib/buckets/buckets.js
danmcd: Thank you!

lib/common.js
danmcd: Thank you!  Resolving.

lib/s3-routes.js
danmcd: Resolving.

lib/buckets/buckets.js
danmcd: Thank you for the fix.  Resolving.

lib/buckets/objects/get.js
danmcd: Thank you for the fix.  Resolving.

danmcd (commented):
Thanks for the fixes.

I'd like to take one more look at s3-multipart.js and s3-routes.js before I approve.

lib/s3-multipart.js
```diff
@@ -639,6 +640,564 @@ function uploadPartWithRecord(req, res, next, uploadRecord, bucketName,
...
+                    return;
+                }
+
+                // Step 3: Parse range
+                var rangeHdr =
+                    req.headers['x-amz-copy-source-range'];
+                var range = parseCopySourceRange(
+                    rangeHdr, srcMeta.content_length);
+
+                if (rangeHdr && !range) {
```
danmcd, line 773: What happens if rangeHdr is null? Looks like you'll just proceed on to line 784 and dereference a null range.

Should this be `if (!range)` ?

Was this your error or Claude's?

danmcd (changes requested):
Check for more antipatterns like this one, please?

lib/s3-multipart.js
cneira: I examined the code again, and I think is ok, this truth table describes the posible outcomes
for line 774.

| rangeHdr | range | rangeHdr && !range | Result |
|----------|-------|--------------------|--------|
| undefined| null |  false  | Full copy using srcMeta.content_length  |
| null     | null |  false  | Full copy using  srcMeta.content_lenght |
| ''     | null |  false  | Full copy object using srcMeta.content_length|
| 'bytes=0-9'   | {length: 10}   | false  |  copy using range|
| 'bytes=0-A'   | null   | true  |  return HTTP 400 InvalidArgument |
| 'other not parsable string'   | null   | true  |  return HTTP 400 InvalidArgument |

The ternary operator in line 784 guards us from dereferencing a null range, as it
will just use srcMeta.content_length.

This code was generated by Claude, reviewed by me.

danmcd: Wow... that's non-obvious, or more-likely I'm just thick.

So two simpler questions, one I'll ask above, and one I'll ask below. Resolving this one.

lib/s3-multipart.js
```diff
@@ -639,6 +640,564 @@ function uploadPartWithRecord(req, res, next, uploadRecord, bucketName,
...
+
+                var copyLen = range ?
+                    range.length :
+                    srcMeta.content_length;
+
+                log.debug({
+                    srcBucket: srcBucket,
+                    srcObject: srcObject,
+                    srcSize: srcMeta.content_length,
+                    range: range,
```
danmcd: So what happens if `range` is null HERE?  Later below (and I'll cite it as a comment, you have this:

`range: srcOpts.range || 'full',`

Should THAT be here instead of possibly passing on a NULL `range` ?

lib/s3-multipart.js
```diff
@@ -639,6 +640,564 @@ function uploadPartWithRecord(req, res, next, uploadRecord, bucketName,
...
+        };
+        if (range) {
+            srcOpts.range = 'bytes=' + range.start +
+                '-' + range.end;
+        }
+
+        log.debug({
+            attempt: attempt,
+            shark: srcShark.manta_storage_id,
+            range: srcOpts.range || 'full',
```
danmcd, line 943: You are checking for null `srcOpts.range` here.  This is what prompted me to ask the question above.  This comment will be resolved in lockstep with the one above.

lib/s3-multipart.js
```diff
@@ -639,6 +640,564 @@ function uploadPartWithRecord(req, res, next, uploadRecord, bucketName,
...
+        return (null);
+    }
+    var m = rangeHeader.match(/^bytes=(\d+)-(\d+)$/);
+    if (!m) {
+        return (null);
+    }
+    var start = parseInt(m[1], 10);
+    var end = parseInt(m[2], 10);
+    if (start > end || end >= sourceSize) {
+        return (null);
```
danmcd, line 662: If some jerk feeds something that fails HERE, we will just instead copy the whole object, even if, maybe, we have copied some properly formed parts?

Imagine:

```
Total object size of 1500...

bytes=0-999
bytes=1000-1199
bytes=1200-1600
```
That last one will return NULL here, so you then perform a copy of 0-1500 even after the first two blocks?

Also, I imagine someone could be much MUCH more obnoxious about what they feed here, unless there's some existing MPU logic (perhaps in streamPartCopy() or something it calls) I've forgotten about?

lib/s3-routes.js
```diff
@@ -2312,6 +2342,71 @@ function getObjectTagsFromProperties(req, res, callback) {
...
+    if (req.query && req.query.location === undefined) {
+        return (false);
+    }
+
+    // We need to bypass the S3 response formatter
+    // as this is a canned response.
+    req._skipS3ResponseProcessing = true;
+    // TODO:
+    // In multi-DC, make this to be populated through SAPI
+    var region = 'us-east-1';
```
danmcd, line 2378: The hardcoded "region" here is part of the "S3 response formatter bypass", right?

danmcd (commented):
Sorry if I'm still being thick here.

lib/s3-routes.js
cneira: That's correct, we  just generated a canned XML with the hardcoded region that's the reason we need to set _skipS3ResponseProcessing to true. That bypasses the JSON to XML conversion from Manta responses to the XML that S3 clients can parse.

lib/s3-multipart.js
cneira: You are right, I have modified this to only do the full copy of the object when the rangeHdr is not present, if the range requested is invalid, then return a 400 or 416.

lib/s3-multipart.js
cneira: I understand here that there is little value to print null in the logs. Maybe it's clearer to print 'full' in  debug logs when we are processing a full copy. That will be  consistent on how range printing is handled in the rest of the code.

lib/s3-multipart.js
danmcd: Resolving.

test/test-scripts/s3-uploadpartcopy-range-test.py
```diff
@@ -0,0 +1,372 @@
...
+    print('=' * 60)
+
+    # Cleanup
+    cleanup(s3)
+
+    sys.exit(0 if failed == 0 else 1)
+
+
+if __name__ == '__main__':
+    main()
```
danmcd, line 796: Like this test so far.  Unless I missed you having these already, or the MPU is strict about sequential parts (VERY sorry if I missed this being the case), consider variants of this in your rclone/MPU/etc testing:

Given a 1MiB block object, consider these crazy MPUs:

OVERLAPPING PARTS:  0-512KiB, 256KiB-786KiB, 786KiB-(1MiB-1)

BROKEN MIDDLE: 0-512KiB, 512KiB-(1MiB+1)

OUT OF ORDER BUT OTHERWISE GOOD: 256KiB-(786KiB-1), 786KiB-(1MiB-1), 0-(256KiB-1)

OUT OF ORDER WITH OVERLAP: 256KiB-786KiB, 786KiB-(1MiB-1), 0-256KiB

lib/s3-multipart.js
danmcd: I put examples of, "MUCH more obnoxious," in your new test script. Resolving this one.

lib/s3-routes.js
danmcd: Resolving.

lib/s3-multipart.js
danmcd: Resolving.

danmcd (commented):
One last question in the tests.

test/test-scripts/s3-uploadpartcopy-range-test.py
```diff
@@ -354,6 +354,430 @@ def main():
...
+    #
+    # Uses a 20MiB source object.  Three parts with deliberate overlaps:
+    #   Part 1: bytes=0-7340031            (0 to 7MiB-1,   7MiB)
+    #   Part 2: bytes=5242880-15728639     (5MiB to 15MiB-1, 10MiB,
+    #                                       overlaps part 1 by 2MiB)
+    #   Part 3: bytes=14680064-20971519    (14MiB to 20MiB-1, 6MiB,
+    #                                       overlaps part 2 by 1MiB)
+    #
+    # Each non-last part is >= 5MB so CompleteMultipartUpload should succeed.
+    # The assembled object = src[0:7M] + src[5M:15M] + src[14M:20M] = 23MiB.
```
danmcd, line 684: 23?!  So 20 source, split with overlap, and I get 23 back?!?  That seems wrong.  SHouldn't I get 20 back to match the source object?  Or is this how S3 WORKS?!?

danmcd (commented):
The successful overlap case (send 23 MB of overlapping segments of a 20MB source) returns the parts EXACTLY?  Is it up to the client to do the right thing?  Does S3 (mis)behave this way?

cneira:
> The successful overlap case (send 23 MB of overlapping segments of a 20MB source) returns the parts EXACTLY? Is it up to the client to do the right thing? Does S3 (mis)behave this way?

It is the responsibility of clients to split an object into parts and assign the correct order to those parts when uploading. The reason is that there is no correspondence between a part index and its content; the integrity check is performed per part.

In a multipart upload, we provide the client with the means to partition an object. The assembly on the server side is then performed by concatenating the parts in part-number order, as sent by the client.

We follow AWS S3 constraints, where integrity validation is done only on a per-part basis. There is no size validation of the final assembled object. The parts are treated as opaque blob objects by the backend, so a client could specify, as done here, overlapping segments or arbitrary size ranges from the source object. This can cause the destination object to differ in size and content from the original object. 

Here are the documents relevant for this topic :  

- https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html

test/test-scripts/s3-uploadpartcopy-range-test.py
cneira: We follow AWS S3 constraints, where integrity validation is done only on a per-part basis. There is no size validation of the final assembled object. The parts are treated as opaque blob objects by the backend, so a client could specify, as done here, overlapping segments or arbitrary size ranges from the source object. This can cause the destination object to differ in size and content from the original object.

test/test-scripts/s3-uploadpartcopy-range-test.py
danmcd: Resolving.

test/test-scripts/s3-uploadpartcopy-range-test.py
danmcd: Resolving.  So this *is* how S3 works too.  Eeesh.

danmcd (dismissed):
Man that overlap behavior bugs me, but if that's how S3 wants it, it's sure not a blocker of your version of it.

travispaul (dismissed):

danmcd (approved):
Good fixes all, sorry I didn't catch them, but this is why one likes to have more than one reviewer and/or self-scrutiny.

travispaul (approved):

PR #72, opened 2026-04-28
MANTA-5519 Bucket scope enforcement middleware

cneira:
This change allows the creation of access keys that grant access to a subset of specific buckets with defined permissions.
The intent is that the account owner or administrator can create access keys that operate only on a specified list of buckets. Unlike STS, these access keys are permanent, and the permission model for bucket operations is simpler (read, read-write, full).

One thing to consider is that STS is still available. A user with bucket-scoped access key could call STS to generate temporary credentials and define roles or policies as supported by the STS endpoint. This adds an extra layer of complexity, as we must ensure that the scope of the original account is respected when using STS.

To create a bucket-scoped access key, use the following JSON payload with the CloudAPI access key endpoint:
```json
{
  "version": 1,
  "permissions": [
    { "bucket": "app-data", "level": "readwrite" },
    { "bucket": "logs-*", "level": "read" }
  ]
}
```
Where permission is an array of buckets with their associated permissions level.
This change needs modifications to the following services to be aware of this new bucket scoped keys:

- mahi
- sdc-ufds
- sdc-cloudapi
- node-mahi
- node-ufds  

 I have used the branch name accesskey-per-bucket in each repo to reflect this. 
 
 Co-Authored-By: Claude Opus 4.6 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)


docs/bucket-scoped-keys.md
```diff
@@ -0,0 +1,330 @@
...
+
+# Bucket-Scoped Access Keys
+
+Bucket-scoped access keys restrict S3 operations to a
+declared set of buckets with per-bucket permission levels.
+A scoped key can only access the buckets named in its scope
+and only at the granted permission level.  An unscoped key
+(the default) has unrestricted access to all buckets.
+
+Scoped keys are managed through CloudAPI and enforced by
```
danmcd, line 14: Dumb question:  Does the new Rust Triton cloudapi client ( https://github.com/TritonDataCenter/monitor-reef/tree/main/cli/triton-cli ) need changes too?

danmcd (commented):
Checkpoint, resuming at lib/auth/credential-provider.js.

docs/bucket-scoped-keys.md
cneira: Both the node and Rust clients need to expose this new option when creating access-keys. I need to create a PR for that, thank you for bringing this up.

danmcd: Thank you, resolving.

lib/server/formatters.js
```diff
@@ -167,12 +167,17 @@ function formatJSON(req, res, body) {
...
+    // always be serialized.  Without this, a scope deny
+    // on a GET/PUT object path passes the raw Error object
+    // to res.write(), crashing Node v6 with
+    // "First argument must be a string or Buffer".
     if (req && (req._skipS3ResponseProcessing ||
     req._binaryUpload || req._binaryOperation)) {
         return function passthroughFormatter(innerReq, innerRes, body) {
-            // Return body completely unchanged -
-            // no formatting at all
+            if (body instanceof Error) {
```
danmcd, line 178: Is this a bugfix independent of the new functionality? Or is it needed only BECAUSE of the new functionality?

You may have said this in the ticket (and if you did I apologize) but if you didn't you should mention if there are some independent bugfixed picked up along with way with this new feature.

danmcd (commented):
Mostly done with pass 0/1, overlooked tests, and if you want me to dig deeply into any of them please let me know (the new BIG files are probably most worth looking at).

lib/server/formatters.js
cneira: This was a latent bug that never happened before this new functionality, it was flawed from the start  as it was not handling Error objects.

danmcd: Thank you!  Resolving.

lib/s3-multipart.js
```diff
@@ -55,8 +55,10 @@ var common = require('./common');
 var constants = require('./constants');
 var copyModule = require('./buckets/objects/copy');
 var createObjectModule = require('./buckets/objects/create');
-var InvalidDurabilityLevelError = require('./errors').
-    InvalidDurabilityLevelError;
+var appErrors = require('./errors');
+var InvalidDurabilityLevelError = appErrors.InvalidDurabilityLevelError;
+var AccessDeniedByKeyScopeError = appErrors.AccessDeniedByKeyScopeError;
```
travispaul: I'm not seeing where `AccessDeniedByKeyScopeError` was used after being added.

lib/auth/bucket-scope.js
```diff
@@ -0,0 +1,410 @@
...
+    /* segments[0] is empty string before leading / */
+    return (segments[1] || null);
+}
+
+/**
+ * @brief Collect list of bucket names allowed by scope
+ *
+ * Used for ListBuckets filtering. Returns an array of
+ * all bucket patterns in the scope, plus a flag for
+ * whether any pattern uses wildcards.
```
travispaul: I'm not seeing where this returns a flag, it appears to be returning an array of bucket names.

travispaul (commented):
Just a couple of minor nits so far from first thorough pass.

Also please version bump package.json.

cneira:
> Just a couple of minor nits so far from first thorough pass.
> 
> Also please version bump package.json.

package.json version has been bumped up.

lib/common.js
```diff
@@ -983,8 +984,10 @@ function sharkStreams(req, res, next) {
     // For AWS chunked uploads, account for encoding overhead
     var maxBytes = req._size;
     var isAwsChunked = req._awsChunkedMPU || // Support multipart uploads
-        // AWS CLI uses content-encoding: aws-chunked
-        (req.isS3Request && req.header('content-encoding') === 'aws-chunked');
+        // Detect aws-chunked via either Content-Encoding: aws-chunked
+        // or x-amz-content-sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD
+        // (the SigV4 streaming signal used by mc/SDKs).
+        (req.isS3Request && awsChunkedEncoding.isAwsChunked(req.headers));
```
danmcd, line 990: This is not the first time I've noticed that `content-encoding` seems to be not as reliable/required as a spec might make it out to be.  Good catch.

cneira: According to the spec here https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html is Content-Encoding: aws-chunked not optional in this case, but seems clients like mc omit this header. Now the code takes into account that situation, the one that's enforced by SigV4 and clients cannot omit is is x-amz-content-sha256.

danmcd: I see you removed it, but your explanation here covers why.  Thank you.

danmcd (commented):
Incremental checkpoint:  Thank you for the cleanup and extra test.

Will run through this again in the code paths primarily.

danmcd (commented):
Checkpoint, just lib/auth/bucket-scope.js left.  Most of these changes seem okay to me.  You will want @travispaul to be satisfied as well. IF I have anything left, it'll be only in the aforementioned file.

danmcd (approved):
Nothing SCREAMING at me; approving.

